PR #2, opened 2016-09-12
Enable Fortran bindings on OSX

astrofrog:
This is a work in progress and isn't ready for review

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -19,14 +19,20 @@ build:
 requirements:
     build:
         - perl
-        - gcc  # [linux]
+        - gcc  # [not win]
     run:
+        - libgfortran  # [osx]
         - libgcc  # [linux]
 
 test:
+
+    requires:
+        - gcc  # [not win]
+
     files:
         - tests/helloworld.c
         - tests/helloworld.cxx
+        - tests/helloworld.f90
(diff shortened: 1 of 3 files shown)


astrofrog:
@ocefpaf @bekozi @minrk @msarahan - this is ready for review. Note that I added but commented out the C++ tests for now since I had some issues with getting them to pass on MacOS X so I think that indicates another issue we can address in a separate PR.

recipe/meta.yaml
```diff
@@ -19,14 +19,20 @@ build:
 requirements:
     build:
         - perl
-        - gcc  # [linux]
+        - gcc  # [not win]
```
minrk: Since build is skipped entirely on Windows, should this be unconditional?

astrofrog: I was told for other recipes to include this for future-proofness, in case we do enable Windows some day, but happy to remove if people prefer.

minrk: Works for me.

minrk:
👍 from me. Can do the CXX tests in another PR.

ocefpaf:
Thanks @astrofrog for the PR and @minrk for the review!

PS: I had issue with CXX too. That is why I left it out :grimacing:

PR #3, opened 2016-09-13
Enable C++ bindings

astrofrog:
The C++ bindings were not enabled

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -13,7 +13,9 @@ fi
 
 
 ./configure --prefix=$PREFIX \
-            --disable-dependency-tracking
+            --disable-dependency-tracking \
+            --enable-mpi-cxx \
+            --enable-mpi-fortran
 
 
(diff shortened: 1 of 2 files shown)


recipe/build.sh
```diff
@@ -13,7 +13,9 @@ fi
 
 
 ./configure --prefix=$PREFIX \
-            --disable-dependency-tracking
+            --disable-dependency-tracking \
+            --enable-mpi-cxx \
+            --enable-mpi-fortran
```
astrofrog: I added the enable fortran flag to make sure it is explicit (but enabled is the default if a fortran compiler is present, which it needs to be here)

astrofrog:
@ocefpaf @bekozi @minrk @msarahan - this is ready for review!

ocefpaf:
Awesome! Thanks @astrofrog!!

astrofrog:
If only we could move this fast with HDF5! ;)

PR #5, opened 2016-11-29
Update to version 2.0.1

moble:


minrk:
The feedstock needed updating, but it should be okay, now. Can you rebase and push again?

moble:
@minrk Thanks for clearing that up!

minrk:
Thanks!

PR #10, opened 2017-02-08
skip mpiexec tests on Linux [travis skip]

### LINKED ISSUE #8: openmpi [linux]
I see that your `meta.yaml` include `linux`, but I cannot find it in `conda-forge`.  Is there a way to build `openmpi` in `linux`

minrk:
Due to a bug in conda-forge setup, mpiexec can't be called during tests because it will prevent upload of the finished package.

When https://github.com/conda-forge/conda-smithy/pull/337 is merged, we should be able to resume calling mpiexec in the tests.

closes #8

### DIFF
--- a/recipe/run_test.sh
+++ b/recipe/run_test.sh
@@ -13,8 +13,14 @@ command -v mpif90
 mpif90 -show
 
 command -v mpiexec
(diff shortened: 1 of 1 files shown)


minrk:
The thing to check here is if the `upload_or_check_existence` script is called after the build.

PR #15, opened 2017-09-26
Bump version, minor fixes, and rerender

dalcinl:


dalcinl:
@minrk This one is ready merge.

recipe/meta.yaml
```diff
@@ -31,9 +31,9 @@ test:
         - tests/helloworld.cxx
         - tests/helloworld.f
         - tests/helloworld.f90
-    commands:
-        - conda inspect objects openmpi  # [osx]
-        - conda inspect linkages openmpi  # [not win]
+#   commands:
+#       - conda inspect objects openmpi  # [osx]
```
minrk: Did you verify that these fail? I believe it's a specific bug in mpich that causes O_NONBLOCK problems.

dalcinl: No, sorry, you are right. Do you think I should revert these changes?

dalcinl: Never mind. I'm going to strip this commit.

dalcinl:
@minrk I stripped the commit removing the `conda inspect` tests. Let's wait for the builds, if everything is fine, I'll merge.

dalcinl:
@minrk I will delay the merge of this PR until I can give it a check to the macOS builds in my laptop. The output of `conda inspect objects openmpi` shows rpath entries that seems unnecessary, as `@loader_path`  entries should be enough to resolve libraries. Thus, I would like to play with the `build.sh` script, maybe we can get better binaries that do not need rewrites for relocation at package install time.

PR #28, opened 2018-10-08
try building back on travis

minrk:
downstream builds have started failing after the switch to circleci.

The main difference is that the circle builds use xcode 9 + mac sdk 10.12, while Travis runs on 10.10. It's possible that the 10.9 SDK isn't fully selected, or there's something truly incompatible with 10.9 when building with newer compilers. I'm more inclined to think that it's not actually building right, though.


minrk:
Nevermind, after setting `-isysroot` in #29, that's enough for it to be truly compatible. No need to rolback the entire build environment.

PR #41, opened 2019-06-29
Update recipe (build and testing)

dalcinl:
Checklist
* [x] Used a fork of the feedstock to propose changes


recipe/meta.yaml
```diff
@@ -1,44 +1,41 @@
...
     requirements:
       build:
-        - perl 5.26.2
-      host:
-        - {{ compiler('fortran') }}
         - {{ compiler('c') }}
         - {{ compiler('cxx') }}
+        - {{ compiler('fortran') }}
         - make  # [unix]
+        - perl 5.26.2
```
minrk: perl is not a host dependency. This should go back to build

recipe/run_test.sh
```diff
@@ -36,24 +34,14 @@ if [[ $PKG_NAME == "openmpi-mpicxx" ]]; then
...
   command -v mpifort
   mpifort -show
 
-  command -v mpif90
-  mpif90 -show
-
-  command -v mpif77
-  mpif77 -show
-
-  mpif77 $FFLAGS $LDFLAGS helloworld.f -o helloworld_f
```
minrk: Why these changes? Should these paths not be tested?

recipe/meta.yaml
dalcinl: Of course! Look again to the diff, note the line below this one: `hosts = []`.

recipe/run_test.sh
dalcinl: In Open MPI,  compiler wrappers (C/C++/Fortran) are symlinks to `opal_wrapper`. As I explained in the MPICH update PR, `mpifort` is the "modern" name for the Fortran compiler wrapper. This is a good move, newer Fortran standards are quite backward compatible with previous one, there is no point on keep adding compiler wrappers named `mpif95`, `mpif03`, `mpif08` and so on, just use `mpifort` from now on. Let's follow best practices, it is really annoying seeing `mpirun` still used after 30 years of `mpiexec` being standardized in MPI-2, so we should stop using `mpif77`/`mpif90` and move on.

Using just `mpifort` for testing both f77 and f90 sources looks enough to me. If you do not agree, I can add additional tests for `mpif77`/`mpif90`, though IMHO these extra tests would provide very little value.

ax3l:
@conda-forge-admin,  please rerender

(so we get the fixed conda-build 3.18.7 with https://github.com/conda/conda-build/pull/3596 )

ax3l:
sorry, still too old in the bot

recipe/meta.yaml
minrk: For history, compilers were placed in host to fix references to compiler paths in 630319f2db912205a84613c60ecc6d1727360331 . I believe this is made obsolete by 1015075f4aa5995e836921db2aa62f249911ba8b so this change should be okay.

minrk:
Pretty sure this will succeed after the aarch64 build finishes. Then happy to have it merged.

dalcinl:
@minrk Please look at the aarch64 build log, particularly the testing of the `openmpi` package. 

```
The following NEW packages will be INSTALLED:

    mpi:     1.0-openmpi      conda-forge
    openmpi: 4.0.1-hc99cbb1_2 local      
```

This smells like dependencies on the C++/Fortran runtime libraries are still broken. Or I am missing something?

Should we just ignore this issue and merge?

I'm tempted to backport the `mpivars` program from MPICH and install it in `$PREFIX/bin`, this way we have an actual binary executable that uses MPI, so we can test the core `openmpi` package. And as a side benefit, the `mpivars` program is actually useful.

minrk:
You're right, https://github.com/conda/conda-build/issues/3583 is indeed *not* fixed in conda-build 3.18.7

minrk:
We still need to wait for conda-build to fix run_exports before we can make any aarch64 packages. This should put a pause on any/all aarch64 builds across conda-forge, I think.

ax3l:

mariusvniekerk:
@minrk Is this right now?

minrk:
@marius triggering rebuild now to check. The last build was not correct.

minrk:
@mariusvniekerk No, the build is still incorrect on aarch64, missing strong run_exports from the compilers.

minrk:

I think the latest CB release may fix this.

mariusvniekerk:
Fingers crossed

dalcinl:

dalcinl:
@beckermr I checked carefully the pkg dependencies of the macOS builds and both look good to me. Could you please double check them and confirm? 

PS: I still need to wait for the Linux aarch64 build to check they are OK (we had dependency issues in there).

beckermr:
LGTM! The aarch64 build will take a while.

beckermr:
@dalcinl Please close #45 when you merge this.

beckermr:
huzzah!

PR #53, opened 2019-12-05
Remove C++ bindings

minrk:
Based on discussion in https://github.com/open-mpi/ompi/pull/7169 it seems that openmpi maintainers view the c++ bindings as deprecated by a decade and plan for its removal in 5.x, so maybe we shouldn't package it.

Is anyone using the mpi c++ bindings? cc @astrofrog @isuruf

Reverts #3

### DIFF
--- a/recipe/build-mpi.sh
+++ b/recipe/build-mpi.sh
@@ -15,6 +15,7 @@ export FC=$(basename "$FC")
 if [ $(uname) == Darwin ]; then
     if [[ ! -z "$CONDA_BUILD_SYSROOT" ]]; then
         export CFLAGS="$CFLAGS -isysroot $CONDA_BUILD_SYSROOT"
+        export CXXFLAGS="$CXXFLAGS -isysroot $CONDA_BUILD_SYSROOT"
     fi
     export LDFLAGS="$LDFLAGS -Wl,-rpath,$PREFIX/lib"
 fi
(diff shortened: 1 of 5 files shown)


recipe/build-mpi.sh
```diff
@@ -15,7 +15,6 @@ export FC=$(basename "$FC")
 if [ $(uname) == Darwin ]; then
     if [[ ! -z "$CONDA_BUILD_SYSROOT" ]]; then
         export CFLAGS="$CFLAGS -isysroot $CONDA_BUILD_SYSROOT"
-        export CXXFLAGS="$CXXFLAGS -isysroot $CONDA_BUILD_SYSROOT"
```
dalcinl: @minrk I would keep this line.

recipe/meta.yaml
```diff
@@ -26,7 +26,6 @@ outputs:
     requirements:
       build:
         - {{ compiler('c') }}
-        - {{ compiler('cxx') }}
```
dalcinl: I think you still need this for configure to run and to support the `mpicxx` compiler wrapper.

astrofrog:
I'm not using the c++ bindings

dalcinl:
@minrk Are we going to remove the C++ bindings in the mpich package as well?

minrk:
I don't know. I'm not aware of the mpich maintainers having such a negative view of their own C++ bindings, and I haven't seen similar issues with it. My inclination would be to leave it alone for now unless someone has a reason to remove them.

xylar:
It seems like any packages that use C++ and OpenMPI are now going to need to be rebuilt.  Otherwise, there are errors like:
```
ESMF_RegridWeightGen: error while loading shared libraries: libmpi_cxx.so.40: cannot open shared object file: No such file or directory
```
See https://github.com/conda-forge/esmf-feedstock/issues/50

Issue #42, opened 2019-07-03
Allow extra configuration arguments 


cjnolet:
It would be nice if extra configuration arguments could be specified somehow. For instance, if we could specify “—with-cuda —with-ucx” so that a CUDA-aware MPI could be built.

jakirkham:
FWIW have put an initial UCX package forward in PR ( https://github.com/conda-forge/staged-recipes/pull/9162 ), which should help here. This still doesn't handle some things that we would like it to, but it is a good foundation for adding that functionality.

jakirkham:
It's probably worth looking at this more closely as we now have a UCX package and are able to perform GPU builds. Thoughts on this @conda-forge/openmpi ?

jakirkham:
@minrk, do you have thoughts here? Happy to do a PR if this is reasonable. 🙂

minrk:
Seems totally reasonable. Happy to have variants if they are useful.

jakirkham:
Thanks Min. I'll work on a PR so we can discuss further.

cjnolet:
This will be great! Currently, we have an MPI communicator implementation to run our automated GTests on RAPIDS cuML but it needs to be disabled by default because the users need to install CUDA-aware manually.

leofang:
@cjnolet @jakirkham Do you happen to have a work in progress here? If not, I might be able to find some time and try enabling a GPU build. I just had an offline discussion with @dalcinl today, and then found this issue matching exactly what I intended to do! 

Context: We will make an official release of mpi4py soon, with the support of `__cuda_array_interface__` included (it was previously considered experimental and only available on master). So, it seems natural to me if we could build a CUDA variant for the Open MPI package, on which the updated mpi4py can then depend.

But, with some quick digging I found that neither OMPI nor mpi4py links to CUDA runtime (on conda-forge it'd be contained in `cudatoolkit` I think). The linking to CUDA is [done on the OMPI side at runtime via dlopen](https://github.com/open-mpi/ompi/blob/d14d8ad30d86f6650cb6c6384ddef304d408ff15/opal/mca/common/cuda/common_cuda.c#L309) (I suppose), and mpi4py doesn't care at all. So, @dalcinl suggested -- I hope I understood him correctly -- we take this route:
[...]

PR #54, opened 2020-01-22
Enable CUDA awareness by default

leofang:
Checklist
* [x] Used a [fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)

Address part of #42. ~~Still a work in progress for testing what kinds of build-time errors we may stumble into.~~ UCX is not enabled yet.

***EDIT***: After a long discussion back and forth in this PR and the parent issue #42, we converged to a much simpler solution: on `linux64`, we build with CUDA 9.2, and remove the `run_exports` dependency on `cudatoolkit` to make it work with both CPU (`cudatoolkit` does not have to present) and GPU (`cudatoolkit` is installed separately). This solution is enabled by two facts:
1. Open MPI always has its CUDA-related build time flags set to 1 for fairly recent CUDA versions
2. Open MPI does not link to CUDA runtime (CUDA awareness kicks in at runtime via successful `dlopen` and `dlsym`)


leofang:

leofang:

leofang:
The re-rendering bot is very unhappy today...Let's re-render later.

leofang:

leofang:

leofang:

leofang:

leofang:
CI passed. For @cjnolet @jakirkham: Would you like to tackle the UCX dependency in this PR, or you'd like to do it separately?

recipe/build-mpi.sh
```diff
@@ -31,7 +39,9 @@ export LIBRARY_PATH="$PREFIX/lib"
             --with-wrapper-cxxflags="-I$PREFIX/include" \
             --with-wrapper-fcflags="-I$PREFIX/include" \
             --with-wrapper-ldflags="-L$PREFIX/lib -Wl,-rpath,$PREFIX/lib" \
-            --with-sge
+            --with-sge \
+            $build_with_cuda
 
 make -j"${CPU_COUNT:-1}"
 make install
+echo opal_warn_on_missing_libcuda = 0 >> $PREFIX/etc/openmpi-mca-params.conf
```
jakirkham: Is this still needed if we are only building with CUDA in certain cases?

recipe/build-mpi.sh
```diff
@@ -20,6 +20,14 @@ if [ $(uname) == Darwin ]; then
     export LDFLAGS="$LDFLAGS -Wl,-rpath,$PREFIX/lib"
 fi
 
+# At this point, we selected the correct docker image, but we have no
+# way to know tell if CUDA is present, so we look for nvcc
+if [ $(command -v nvcc) ]; then
```
jakirkham: Conda-build passes through all the variants into the build environment. So we can just check this directly.

```suggestion
if [ ${cuda_compiler_version} != "None" ]; then
```

jakirkham:
> CI passed. For @cjnolet @jakirkham: Would you like to tackle the UCX dependency in this PR, or you'd like to do it separately?

That would be useful, but I don't think that should hold up this PR.

Also it would be good to double check what compatibility guarantees UCX provides and add this to either [conda-forge-pinning]( https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/master/recipe/conda_build_config.yaml ) or [the ucx package]( https://github.com/conda-forge/ucx-split-feedstock ) before having other things link to it. Have raised an issue about adding this to the ABI tracker we use ( https://github.com/lvc/abi-tracker/issues/12 ).

recipe/build-mpi.sh
dalcinl: @leofang Follow @jakirkham advice, mess with the config var ONLY if `$build_with_cuda` is not the empty string `if [ -z "$build_with_cuda" ] ...`

jakirkham: Thanks for confirming Lisandro! I wasn't actually sure how OpenMPI used this.

For more context `cudatoolkit` automatically becomes a dependency when `{{ compiler("cuda") }}` is added. So would expect this isn't needed in that case either and can just be dropped.

Though you know more about OpenMPI than I do. So please feel free to correct me if needed 🙂

dalcinl: > Thanks for confirming Lisandro! I wasn't actually sure how OpenMPI used this.
> 

Setting it unconditionally would make no harm, but just in case, better safe than sorry.

> For more context `cudatoolkit` automatically becomes a dependency when `{{ compiler("cuda") }}` is added. So would expect this isn't needed in that case either and can just be dropped.
> 

Yes, @leofang confirmed that Open MPI access the CUDA runtime via `dlopen()` at runtime, thus being optional, so the CUDA runtime **SHOULD NOT** become a dependency of the openmpi package. However, I have no idea how to force `conda` to remove that implicit dependency. Any tips?

jakirkham: Just for my own understanding, what is the motivation for not having a `cudatoolkit` dependency?

leofang: Yes, I found that too. Seems like `conda install openmpi` would also install `cudatoolkit`...Would be preferable not to.

leofang: > Just for my own understanding, what is the motivation for not having a `cudatoolkit` dependency?

Many MPI applications do not need GPU.

jakirkham: Thanks for the context. When adding `{{ compiler("cuda") }}`, you will get 4 builds of `openmpi`. 3 with different CUDA versions and 1 that does not use CUDA. The one thing this recipe doesn't have right now is a way to select between the CUDA and non-CUDA installs. I'll make a suggestion in the diff as to how we can add that.

leofang: @jakirkham maybe we can do this?
```
ignore_run_exports:
  - cudatoolkit
```

jakirkham:
To better aid in the selection between CPU and GPU packages would make the following changes to the recipe. These would create an `openmpi-proc` package. Users can then choose CPU or GPU variants by installing either `openmpi-proc=*=cpu` or `openmpi-proc=*=gpu`. Please let me know if you have any questions. Happy to discuss further 🙂

Would add the following changes to `recipe/meta.yaml`:

This Jinja can be added near the top:

```
{% set openmpi_proc_type = "cpu" if cuda_compiler_version == "None" else "gpu" %}
```

This can be added to the `outputs` section:

```yaml
outputs:
  - name: openmpi-proc
    version: 1.0.0
    build:
      number: 0
      string: "{{ openmpi_proc_type }}"
    test:
      commands:
        - exit 0
    about:
      home: https://github.com/conda-forge/openmpi-feedstock
      summary: A meta-package to select CPU or GPU OpenMPI build.
  ...
```

Other packages that have a CPU or GPU build should add this to their requirements:

```yaml
    requirements:
      ...
      run_constrained:
        - openmpi-proc * {{ openmpi_proc_type }}
```

recipe/build-mpi.sh
jakirkham: Please see [this comment]( https://github.com/conda-forge/openmpi-feedstock/pull/54#issuecomment-577376632 ). For a suggested way to address this.

jakirkham: Sorry I wrote that comment before I saw yours, Leo.

Unfortunately `ignore_run_exports` isn't likely to do what we want here. In the 1 CPU build `cudatoolkit` is already ignored. In the 3 CUDA builds this will result in packages that have linked to some version of CUDA, but no longer depend on `cudatoolkit`. This means users would need to install this `cudatoolkit`. Further they wouldn't know what version of CUDA packages were built with and would lack a way to select packages that mean their CUDA requirements. This would complicate deployment to a system that expects a specific version of CUDA (due to driver versions or other constraints).

leofang: > Sorry I wrote that comment before I saw yours, Leo.

No worries, John.  I realized that too. 😄 

> Unfortunately `ignore_run_exports` isn't likely to do what we want here. In the 1 CPU build `cudatoolkit` is already ignored. In the 3 CUDA builds this will result in packages that have linked to some version of CUDA, but no longer depend on `cudatoolkit`. This means users would need to install this `cudatoolkit`. 

1. The linking is done in runtime, so technically it might be even possible that `openmpi` built with CUDA 10.2 could work with CUDA 9.2 discovered in runtime. (@jsquyres I am crying out again for confirmation 😭)
2. We do expect users to install `cudatoolkit` explicitly, either by themselves or by the constraint of some other packages like `cupy` or `pytorch`.

> Further they wouldn't know what version of CUDA packages were built with and would lack a way to select packages that mean their CUDA requirements. This would complicate deployment to a system that expects a specific version of CUDA (due to driver versions or other constraints).

If item 1 above is true, this would not be any problem. If not, then we do need a selector or your proposal below to enforce this requirement.

leofang:
> To better aid in the selection between CPU and GPU packages would make the following changes to the recipe. These would create an `openmpi-proc` package. Users can then choose CPU or GPU variants by installing either `openmpi-proc=*=cpu` or `openmpi-proc=*=gpu`. Please let me know if you have any questions. Happy to discuss further 🙂

Thanks John! I suppose this is how the UCX recipe works? I wonder if it could achieve the following things:
1. `conda install openmpi` would by default install the CPU version
2. `conda install openmpi cudatoolkit` (or replacing `cudatoolkit` by any GPU libraries depending on it) would select the right GPU version

Another model that I considered earlier is to following h5py, in which a flag is set in `conda_build_config.yaml`: https://github.com/conda-forge/h5py-feedstock/blob/master/recipe/conda_build_config.yaml (we could have an analog like
```
gpu:
  - nocuda
  - cuda92
  - cuda100
  - cuda101
  - cuda102
```
or something simpler to make a distinction), and the CPU version has a higher build number: https://github.com/conda-forge/h5py-feedstock/blob/050d44e50ee9561049249ae0205f28b6501b2bd6/recipe/meta.yaml#L7-L10
so that it's preferred when a user does not ask for it explicitly.

leofang:
Another worrying thing: this PR generates 5 outputs as discussed above. What happens when we build mpi4py (without any change to the recipe)? Would it also generate 5 variants because of the dependency on `openmpi`?

recipe/build-mpi.sh
jakirkham: Ok thanks for that additional context. Let's wait and see what others from the OpenMPI world say then 🙂

jakirkham:
Yeah setting up a default variant should be possible. Once we figure out how OpenMPI should be handling CUDA versions generally 😉

As to `mpi4py`, that depends on what API it is using. Is it sensitive to CUDA being present or which version of CUDA is used? If so, then we might need to build more variants of `mpi4py`. If not, we can build whatever limited subset are needed to correctly reproduce its functionality. Happy to discuss that more once we get there 🙂

recipe/build-mpi.sh
leofang: Oh cool, thanks @jakirkham. I didn’t know this. It’s much better.

leofang:
Thanks, John. A quick reply:

> As to `mpi4py`, that depends on what API it is using. Is it sensitive to CUDA being present or which version of CUDA is used? If so, then we might need to build more variants of `mpi4py`. If not, we can build whatever limited subset are needed to correctly reproduce its functionality. Happy to discuss that more once we get there 🙂

No, `mpi4py` will not care how `openmpi` is built or whether/which CUDA is present. The CUDA awareness is completely relying on the underlying MPI implementation. So [our plan](https://github.com/conda-forge/openmpi-feedstock/issues/42#issuecomment-577005702) was to make no change to mpi4py-feedstock, and to continue building only 1 variant of `mpi4py` on top of `openmpi` as before. However, since `mp4ipy` has a `pin_run_as_build` clause: https://github.com/conda-forge/mpi4py-feedstock/blob/9c96f7a57d3bdc44dede557dbe8cfa3e1a523d61/recipe/conda_build_config.yaml#L5-L7, it now seems 5 variants of `mpi4py` will be generated with this PR?! This is not I wanted...

I'm aiming to get it right here, since Open MPI is the only MPI implementation on CF that supports CUDA afaik.

recipe/build-mpi.sh
jsquyres: Oops, sorry -- missed this prompting for information -- thanks to @jakirkham for poking me with a pointer to this comment.

> Yes, @leofang confirmed that Open MPI access the CUDA runtime via dlopen() at runtime, thus being optional, so the CUDA runtime SHOULD NOT become a dependency of the openmpi package. 

That is correct.  I haven't read this entire lengthy conversation, but I think that our reasons for doing that were for almost exactly the use case of this PR: we want CUDA to be able to be used, but some clusters only have funding for GPUs on some of their nodes (and therefore don't have the CUDA libraries/etc. installed on all nodes in their cluster).  But they still want a single Open MPI installation to be able to be used on _all_ nodes -- both with and without the CUDA libraries/etc.  So we made some `cuda` components in Open MPI, but they manually try to `dlopen("libcuda.so.1", ...)` at run time, as opposed to having an `ld`-mandated requirement for finding that library at run time.

That's what the `opal_warn_on_missing_cuda` MCA param is for: if you configured Open MPI to expect CUDA support, we warn if we can't find it at run time.  BUT: if you have a scenario like I outlined above (e.g., one Open MPI install that spans nodes with and without CUDA), then it can be ok if Open MPI doesn't find the CUDA library at run time (e.g., if it's running on a node that doesn't have a GPU / doesn't have CUDA).  So you found exactly the right MCA param to shut Open MPI up about this warning.

Does this answer your question?

jsquyres: Also:

> The linking is done in runtime, so technically it might be even possible that openmpi built with CUDA 10.2 could work with CUDA 9.2 discovered in runtime. (@jsquyres I am crying out again for confirmation 😭)

Open MPI is specifically looking for `libcuda.so.1` or `libcuda.dylib`.  So even if you have CUDA 9.2 or CUDA 10.2, if they have `libcuda.so.1`, it should be ok.  I.e., I _assume_ that NVIDIA provides the usual ABI guarantees about SO version numbers.

leofang:
With Jeff's comments (https://github.com/conda-forge/openmpi-feedstock/pull/54#discussion_r370311640, https://github.com/conda-forge/openmpi-feedstock/pull/54#discussion_r370312867), I think this recipe can be made very simple: for `linux64` we always turn on CUDA build and use CUDA compiler, say, 9.2. Then, we use `ignore_run_output` to remove the dependency on `cudatoolkit`. 

Net effect: only 1 flavor is built on `linux64`, and if `cudatoolkit` happens to be installed and present, CUDA awareness will be in effect.

@jakirkham, would the above approach work?

jakirkham:
I'm still [missing the same information]( https://github.com/conda-forge/openmpi-feedstock/issues/42#issuecomment-577852756 ) you are, Leo. Once we know that I think it should help clarify what we should do here. 🙂

leofang:
Since the check suggested by Jeff (https://github.com/conda-forge/openmpi-feedstock/issues/42#issuecomment-577880633) is verified, I took the approach proposed in https://github.com/conda-forge/openmpi-feedstock/pull/54#issuecomment-577869100. @jakirkham PTAL, thanks!

leofang:
Note that as expected, [cuda plugin is built](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=114173&view=logs&j=d0d954b5-f111-5dc4-4d76-03b6c9d0cf7e&t=6d4b912b-175d-51da-0fd9-4d30fe1eb4e7&l=21475), but [`cudatoolkit` is not installed during testing](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=114173&view=logs&j=d0d954b5-f111-5dc4-4d76-03b6c9d0cf7e&t=6d4b912b-175d-51da-0fd9-4d30fe1eb4e7&l=21650).

recipe/meta.yaml
```diff
@@ -14,20 +14,23 @@ source:
...
+        - cudatoolkit
     requirements:
       build:
         - {{ compiler('c') }}
         - {{ compiler('cxx') }}
         - {{ compiler('fortran') }}
+        - {{ compiler('cuda') }}  # [linux64]
         - make  # [unix]
         - perl 5.26.2
       host:
```
jakirkham: Would add the following after `run`. This should make sure that users are not able to install a `cudatoolkit` older than the one we built with.

```yaml
      run_constrained:
        - cudatoolkit >={{ cuda_compiler_version }}
```

leofang:

recipe/meta.yaml
leofang: Thanks, John! Added in aea185b.

jakirkham: cc @isuruf (in case he has a better suggestion on how to add this 😉)

jakirkham:
@cjnolet @minrk, could you please take a look? 🙂

recipe/meta.yaml
isuruf: Does this really work with any CUDA version if built using 9.2?

jakirkham: Yeah that's what we found through experimentation and discussion with OpenMPI devs. Would read [this thread]( https://github.com/conda-forge/openmpi-feedstock/pull/54#discussion_r369764497 ) and [this thread]( https://github.com/conda-forge/openmpi-feedstock/issues/42#issuecomment-577813078 ) for more info.

leofang: Thanks @jakirkham!

leofang:
@dalcinl any comment? Thanks!

dalcinl:
> @dalcinl any comment? Thanks!

Oh, boy... Don't you had enough from me :laughing:?

jakirkham:
Thanks Lisandro! Your feedback has been really helpful 😄

dalcinl:
> Thanks Lisandro! Your feedback has been really helpful

Just to clarify, I was just teasing @leofang. I was really annoying with him in the last mpi4py PR he authored!

leofang:
> I was really annoying with him in the last mpi4py PR he authored!

Thank god this PR is much easier! 🤣

leofang:
@conda-forge/openmpi It'd be nice if one of you could press the green button. Thanks.

dalcinl:
@leofang Merged! Once the new build is available, please try it out and confirm it actually works! I'm a little bit worried that we do not have tests for this new feature, but I understand that it may probe difficult to implement. If that is possible, then we you should consider adding some minimal tests (maybe just a two-process send/recv) making sure that we can really communicate GPU buffers.

leofang:
Thanks, @dalcinl! Yeah it would be nice if the CUDA awareness could get tested; however, currently the condo-forge CIs do not have GPU for us to test, and frankly this is very scary because we have been building GPU packages in the blind, and stupid bugs do happen. 

I am hoping we could set up [gpuCI](https://docs.rapids.ai/gpuci) in the feedstocks to do the job, but I am not sure if it's possible. Any comments, @jakirkham and @gmarkall?

leofang:
> Once the new build is available, please try it out and confirm it actually works! I'm a little bit worried that we do not have tests for this new feature, but I understand that it may probe difficult to implement. If that is possible, then we you should consider adding some minimal tests (maybe just a two-process send/recv) making sure that we can really communicate GPU buffers.

@dalcinl Just did `conda install -c conda-forge openmpi-mpicc`, built `mpi4py` from master, and run tests against CuPy 7.1.1 and Numba 0.48.0 (just released this week). Also tested `mpi4py/demo/cuda-aware-mpi/use_*.py`. Everything works. (Though I did not install `cudatoolkit` from conda-forge and used system CUDA instead.) I think this offers a good confidence for us to release new mpi4py!

jakirkham:
I'd also be curious how this works *with* `cudatoolkit` installed 😉

leofang:
The test with both `cupy` and `cudatoolkit` coming from conda-forge also passed. However, I am not sure which CUDA (conda's or the system's) was discovered at runtime. Any suggestion how to confirm this, @jakirkham?

isuruf:
@leofang, `openmpi` doesn't use any of the libraries coming from `cudatoolkit`. So, it doesn't matter whether it was installed or not.

jakirkham:
Could start a process (like a Python interpreter) and run your code. Then call `lsof -p` on the PID of the interpreter. Should spit out a list of files opened (including libraries loaded).

leofang:
Ah, @isuruf is right -- `openmpi` only needs `libcuda`, not `libcudart`. Sorry for being stupid, I was probably too hungry 😅 

Also thanks to @jakirkham for the useful trick! There's something weird I'm investigating with CuPy's detection on the `CUDA_PATH` env variable. It may be related to cupy/cupy#3006, so we should move the discussion there.

jakirkham:
Good point. Sorry for missing that.

Ok sounds good. Will follow up with you there.

PR #95, opened 2022-04-17
ENH add external build for linking against system MPI libs

### LINKED ISSUE #90: Support for external openmpi libraries
- [x] I read [the conda-forge documentation](https://conda-forge.org/docs/user/introduction.html#how-can-i-install-packages-from-conda-forge) and could not find the solution for my problem there.

Issue:

I am a happy user of `conda install mpich=3=*_external` so I can utilize the mpich libraries available at NERSC (HPC).  I'd like to do the same with openmpi.  Looking at the [doc which mentions similar openmpi support is not yet available](https://conda-forge.org/docs/user/tipsandtricks.html#using-external-message-passing-interface-mpi-libraries), I'm wondering if things have improved with ABI compatibility so that this could be considered?

beckermr:
This PR adds a dummy package with lower solver priority to allow for users at HPC systems to link their packages in conda compiled against openmpi to the system openmpi libraries. This is a requirement for most HPC systems. The other major mpi distribution in conda-forge, `mpich` already supports this. See PR https://github.com/conda-forge/mpich-feedstock/pull/44. 

We'll need to change the docs which mentions this for mpich here: https://conda-forge.org/docs/user/tipsandtricks.html#using-external-message-passing-interface-mpi-libraries

Closes #90 

Checklist
* [x] Reset the build number to `0` (if the version changed)

### DIFF
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -8,8 +8,12 @@ jobs:
     vmImage: ubuntu-latest
   strategy:
     matrix:
-      linux_64_c_compiler_version7cuda_compiler_version10.2cxx_compiler_version7fortran_compiler_version7:
-        CONFIG: linux_64_c_compiler_version7cuda_compiler_version10.2cxx_compiler_version7fortran_compiler_version7
+      linux_64_cuda_compiler_version10.2mpi_typeconda:
+        CONFIG: linux_64_cuda_compiler_version10.2mpi_typeconda
+        UPLOAD_PACKAGES: 'True'
+        DOCKER_IMAGE: quay.io/condaforge/linux-anvil-cos7-cuda:10.2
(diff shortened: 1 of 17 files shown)


beckermr:
@conda-forge-admin rerender

minrk:
Is it possible to get a skip so it isn't rebuilding the same empty package on all platform combinations?

beckermr:
The empty package is not the same since it is on different platforms. I worry the possible difference in solver priority between noarch and platforms would mess with track features. I don't think this is a good idea.

minrk:
Fair enough. I think using `noarch: generic` would be the right way to go for the empty package.

beckermr:
Thanks @minrk. Sorry for the quick merge here. If you’d like to be using noarch here, we can pick that up in a new PR.

minrk:
Not a big deal! The build matrix is just getting big, especially with cuda compilers now. I might have a go the next time I'm bothered waiting for some builds to complete, but waiting-for-build time is the only real cost.

PR #84, opened 2021-03-23
Fix CUDA awareness 

leofang:
Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [x] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [x] Ensured the license file is being packaged.

Close #83. Fix a silly mistake that I made in #80...😞

PR #97, opened 2022-04-21
Use default compiler versions

### LINKED ISSUE #94: openmpi 4.1.3 build links older libgfortran than 4.1.2
### Solution to issue cannot be found in the documentation.

- [X] I checked the documentation.

### Issue

openmpi 4.1.3 is built with gfortran 7, while openmpi 4.1.2 is built with gfortran 9 ([change here](https://github.com/conda-forge/openmpi-feedstock/pull/93/files#diff-dd114495a3204d0d813b8b153a31a968f52a589ced2078a00a13977bea4931bcL24-R24). This results in a runtime dep on libgfortran4 in the newer build instead of libgfortran5, meaning it can't be installed with other packages built with the current default fortran compiler, like hdf5 1.12.

It's pinning the cuda compiler version to 10.2 [here](https://github.com/conda-forge/openmpi-feedstock/blob/fb7ec847d7545ff94d194dfebe283419f4bdb4fc/recipe/meta.yaml#L17) that is the cause, due to zip_keys [here in conda_forge_pinning](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/c01ba229b9c93c4c68f4805e2d312f976e06d1f7/recipe/conda_build_config.yaml#L35-L60).

I'm not sure if there's an easy way in conda_build_config.yaml to specify the old cuda compiler with the new fortran compiler. Perhaps the simplest solution is to remove the pin and just let the compiler matrix do its thing. This is a bit wasteful if we ignore run exports from the cuda compiler, but it should result in correct (if redundant) outputs.

### Installed packages

```shell
mamba create --dry-run -n test hdf5=1.12.1 openmpi=4.1.3 h5py=3.6


(conflict is on libgfortran5 in hdf5 and libgfortran4 in openmpi)
```

### Environment info

```shell
active environment : base
    active env location : /opt/conda
            shell level : 1
       user config file : /root/.condarc
 populated config files : /opt/conda/.condarc
          conda version : 4.12.0
    conda-build version : not installed
         python version : 3.9.10.final.0
       virtual packages : __linux=5.10.104=0
                          __glibc=2.31=0
                          __unix=0=0
                          __archspec=1=x86_64
       base environment : /opt/conda  (writable)
      conda av data dir : /opt/conda/etc/conda
  conda av metadata url : None
           channel URLs : https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
          package cache : /opt/conda/pkgs
                          /root/.conda/pkgs
       envs directories : /opt/conda/envs
                          /root/.conda/envs
               platform : linux-64
             user-agent : conda/4.12.0 requests/2.27.1 CPython/3.9.10 Linux/5.10.104-linuxkit ubuntu/20.04.4 glibc/2.31
                UID:GID : 0:0
             netrc file : None
           offline mode : False
```

minrk:
Because cuda compilers are matrix-aligned with other compilers (c, fortran, etc.), skipping all but an old cuda compiler version means pinning all compilers to old versions. As far as I know, this is only a problem for gfortran, where gfortran 7 emits a dependency on libgfortran 4, while gfortran 10 emits a dependency on libgfortran 5, making their build products mutually exclusive at install time (i.e. #94 means openmpi 4.1.3 cannot be installed along with any other package compiled with the current default fortran on conda-forge).

- specify cuda_compiler_version explicitly instead of via skips
- avoid duplicating version in multiple places, which should prevent future instances of #84 
- some template shenanigans ('cooda') to avoid pulling in the cuda_compiler zip_keys build matrix. I'm not 100% sure this will work. If not, I think we'll have to subset every entry in the compiler zip_keys lists, which will be tedious to maintain.

closes #94

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,11 +1,15 @@
 {% set version = "4.1.3" %}
 {% set major = version.rpartition('.')[0] %}
-{% set build = 1 %}
+{% set build = 2 %}
 
 # give conda package a higher build number
 {% if mpi_type == 'conda' %}
 {% set build = build + 100 %}
 {% endif %}
+
+# avoid conda-forge's cuda compiler detection
+# which forces a complex build matrix
+{% set cooda = 'cuda' %}
 
 package:
   # must not match any outputs for requirements to be handled correctly
@@ -19,8 +23,8 @@ source:
 
 build:
   number: {{ build }}
-  # remember to bump the CUDA version in build-mpi.sh too!
-  skip: true  # [win or (linux64 and cuda_compiler_version != '10.2') or ((aarch64 or ppc64le) and cuda_compiler_version not in (undefined, "None"))]
+  number: 1
+  skip: true  # [win]
 
 outputs:
   {% if mpi_type == 'external' %}
@@ -41,13 +45,13 @@ outputs:
       run_exports:
         - {{ pin_subpackage('openmpi', min_pin='x.x.x', max_pin='x') }}
       ignore_run_exports_from:
-        - {{ compiler('cuda') }}  # [linux64]
+        - {{ compiler(cooda) }}  # [linux64]
     requirements:
       build:
         - {{ compiler('c') }}
         - {{ compiler('cxx') }}
         - {{ compiler('fortran') }}
-        - {{ compiler('cuda') }}  # [linux64]
+        - {{ compiler(cooda) }}  # [linux64]
         - autoconf  # [unix]
         - automake  # [unix]
         - libtool   # [unix]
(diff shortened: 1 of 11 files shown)


recipe/meta.yaml
```diff
@@ -61,7 +65,7 @@ outputs:
         - zlib
         - mpi 1.0 openmpi
       run_constrained:
-        - cudatoolkit >=9.2  # [linux64]
+        - cudatoolkit >={{ cuda_compiler_version }}  # [linux64]
```
minrk, line 68: This was hardcoded as 9.2, and I don't understand the reasoning there. Maybe it should still be 9.2? Or not version-pinned at all?

isuruf:
See https://github.com/conda-forge/openmpi-feedstock/pull/98

PR #104, opened 2022-05-06
noarch external

minrk:
Make external build `noarch: generic`

It's an empty package, so no reason to be platform-specific, I think. Eliminates several CI jobs.


dalcinl:
@minrk After a few restarts, I managed Travis CI to finish. If you are done with this one, please merge.

leofang:
Question: Would this impact the the priority as seen by the conda solver?

beckermr:
Well track features is attached so I think it should not get pulled in unless requested.

recipe/conda_build_config.yaml
```diff
@@ -1,5 +1,5 @@
 mpi_type:
-  - external
+  - external  # [linux and x86_64]
```
leofang, line 2: One more question: Why do we only limit to linux64?

leofang: (still trying to catch up the recent development in this feedstock...)

minrk: just so it builds only once, since the package is the same on all platforms.

jakirkham:
Think this came up before, but we could also move this to the `ucx` side ( https://github.com/conda-forge/ucx-split-feedstock/issues/26 )

recipe/conda_build_config.yaml
leofang: ah ok, I thought by marking `noarch` conda-build or conda-smithy would figure this out, but sounds like we have to hack it ourselves 😅

leofang:
@jakirkham What do you mean by "this"?

jakirkham:
The suggestion of moving things over to `ucx`

recipe/conda_build_config.yaml
minrk: Ah, yes. It would, were it not for the other outputs which _aren't_ noarch. If it were just a single (or all) noarch package, it would figure it out. It's really _how_ we are adding the noarch version via a build matrix that means we have to avoid expanding the cartesian product of `mpi_type` and platform, which is conda-build's standard behavior for multiple values.

PR #133, opened 2023-12-22
recipe: Add libnl dependency on Linux

dalcinl:
Checklist

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -46,6 +46,7 @@ outputs:
         #- prrte
         - libhwloc
         - libevent
+        - libnl  # [linux]
         - zlib
         - ucx  # [enable_ucx]
         - cuda-version {{ cuda_version }}  # [enable_cuda]
@@ -55,6 +56,7 @@ outputs:
         #- prrte
         - libhwloc
         - libevent
+        - libnl  # [linux]
         - zlib
       run_constrained:
         - {{ pin_compatible("ucx", max_pin="x.x") }}  # [enable_ucx]
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -54,10 +54,6 @@ outputs:
         - mpi 1.0 openmpi
         #- openpmix
         #- prrte
-        - libhwloc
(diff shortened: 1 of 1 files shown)


recipe/meta.yaml
```diff
@@ -55,6 +56,7 @@ outputs:
         #- prrte
         - libhwloc
         - libevent
+        - libnl  # [linux]
```
leofang, line 59: Nit: I think all three of them have run_exports set up, so we don't need to add them manually.
```suggestion
```

leofang (approved):

recipe/meta.yaml
dalcinl: I do not understand your suggestion. This is the `run` section, not the `run_constrained` one. We definitely need libnl as a runtime dependency.

dalcinl: The actual problem is: I do not understand this stuff in full, and with all the other related and unrelated mess, I get very very frustrated.

leofang: "run_exports" is used to ensure ABI compatibility by conda-build. If a package is run_export'd, it means when used in the host section, conda-build will also add it to the run section, with proper pinning constraint set up. So we don't have to worry about it and let the upstream recipe maintainers (in this case, libnl et al) decide the pinning for us.

leofang: If a package is not run_export'd, what we do here is wrong. We'd need to use pin_compatible to declare the pinning manually ourselves. It cannot be free dependencies without constraints.

dalcinl:
@leofang What about zlib? Should we perhaps use libzlib in the run section to not depend in the whole zlib package? https://github.com/conda-forge/zlib-feedstock/blob/main/recipe/meta.yaml

leofang:
Took a quick look, we need `zlib` in host, because it offers the headers. But we can remove it from run, because `zlib` run_exports libzlib

leofang (approved):

leofang:
It's just time out, let's merge and retry on main. Thanks, Lisandro!

PR #142, opened 2024-02-05
Rebuild

dalcinl:
Checklist

### DIFF
--- a/conda-forge.yml
+++ b/conda-forge.yml
@@ -15,3 +15,5 @@ provider:
 test: native_and_emulated
 azure:
   free_disk_space: true
+  settings_linux:
+    swapfile_size: 10GiB
--- a/conda-forge.yml
+++ b/conda-forge.yml
@@ -13,3 +13,5 @@ provider:
   linux_ppc64le: default
   osx_arm64: default
 test: native_and_emulated
+azure:
+  free_disk_space: true
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -22,7 +22,6 @@ build:
   number: {{ build }}
   skip: true  # [win]
   skip: true  # [linux and cuda_compiler_version != "11.8"]
-  rpaths_patcher: patchelf  # [linux]
 
 outputs:
   {% if mpi_type == 'conda' %}
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -41,6 +41,7 @@ outputs:
         - {{ compiler('cxx') }}
         - {{ compiler('cuda') }}     # [cuda_compiler != "None"]
         - {{ compiler('fortran') }}
+        - sysroot_{{ target_platform }} 2.17  # [linux]
         #- autoconf  # [unix]
         #- automake  # [unix]
         - libtool   # [unix]
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -52,6 +52,7 @@ outputs:
         - libevent
         - libnl  # [linux]
         - zlib
+        - rdma-core 49  # [linux]
         - ucx  # [linux]
         - cuda-version {{ cuda_compiler_version }}  # [cuda_compiler != "None"]
       run:
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -52,7 +52,6 @@ outputs:
         - libevent
         - libnl  # [linux]
         - zlib
-        - rdma-core 49  # [linux]
         - ucx  # [linux]
         - cuda-version {{ cuda_compiler_version }}  # [cuda_compiler != "None"]
       run:
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -22,6 +22,7 @@ build:
   number: {{ build }}
   skip: true  # [win]
   skip: true  # [linux and cuda_compiler_version != "11.8"]
+  rpaths_patcher: patchelf  # [linux]
 
 outputs:
   {% if mpi_type == 'conda' %}
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,7 +1,7 @@
 {% set version = "5.0.1" %}
 {% set major = version.rpartition('.')[0] %}
 {% set cuda_major = (cuda_compiler_version|default("11.8")).rpartition('.')[0] %}
-{% set build = 2 %}
+{% set build = 3 %}
 
 # give conda package a higher build number
 {% if mpi_type == 'conda' %}
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -35,6 +35,27 @@ jobs:
   timeoutInMinutes: 360
 
   steps:
+  - script: |
+         sudo mkdir -p /opt/empty_dir || true
+         for d in \
+                  /opt/ghc \
+                  /opt/hostedtoolcache \
+                  /usr/lib/jvm \
+                  /usr/local/.ghcup \
+                  /usr/local/lib/android \
+                  /usr/local/share/powershell \
+                  /usr/share/dotnet \
+                  /usr/share/swift \
+                  ; do
+           sudo rsync --stats -a --delete /opt/empty_dir/ $d || true
+         done
+         sudo apt-get purge -y -f firefox \
+                                  google-chrome-stable \
+                                  microsoft-edge-stable
+         sudo apt-get autoremove -y >& /dev/null
+         sudo apt-get autoclean -y >& /dev/null
+         df -h
+    displayName: Manage disk space
   # configure qemu binfmt-misc running.  This allows us to run docker containers
   # embedded qemu-static
   - script: |
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -56,6 +56,12 @@ jobs:
          sudo apt-get autoclean -y >& /dev/null
          df -h
     displayName: Manage disk space
+  - script: |
+         sudo fallocate -l 10GiB /swapfile || true
+         sudo chmod 600 /swapfile || true
+         sudo mkswap /swapfile || true
+         sudo swapon /swapfile || true
+    displayName: Create swap file
   # configure qemu binfmt-misc running.  This allows us to run docker containers
   # embedded qemu-static
   - script: |
--- a/.ci_support/linux_64_c_compiler_version11cuda_compilernvcccuda_compiler_version11.8cxx_compiler_version11fortran_compiler_version11mpi_typeconda.yaml
+++ b/.ci_support/linux_64_c_compiler_version11cuda_compilernvcccuda_compiler_version11.8cxx_compiler_version11fortran_compiler_version11mpi_typeconda.yaml
@@ -36,9 +36,9 @@ zip_keys:
 - - c_compiler_version
   - cxx_compiler_version
   - fortran_compiler_version
+  - cdt_name
   - cuda_compiler
   - cuda_compiler_version
-  - cdt_name
   - docker_image
 zlib:
 - '1.2'
--- a/.ci_support/linux_64_c_compiler_version11cuda_compilernvcccuda_compiler_version11.8cxx_compiler_version11fortran_compiler_version11mpi_typeexternal.yaml
+++ b/.ci_support/linux_64_c_compiler_version11cuda_compilernvcccuda_compiler_version11.8cxx_compiler_version11fortran_compiler_version11mpi_typeexternal.yaml
@@ -36,9 +36,9 @@ zip_keys:
 - - c_compiler_version
   - cxx_compiler_version
   - fortran_compiler_version
+  - cdt_name
   - cuda_compiler
   - cuda_compiler_version
-  - cdt_name
   - docker_image
 zlib:
 - '1.2'
--- a/.ci_support/linux_aarch64_c_compiler_version11cuda_compilernvcccuda_compiler_version11.8cxx_compiler_version11fortran_compiler_version11mpi_typeconda.yaml
+++ b/.ci_support/linux_aarch64_c_compiler_version11cuda_compilernvcccuda_compiler_version11.8cxx_compiler_version11fortran_compiler_version11mpi_typeconda.yaml
@@ -40,9 +40,9 @@ zip_keys:
 - - c_compiler_version
   - cxx_compiler_version
   - fortran_compiler_version
+  - cdt_name
   - cuda_compiler
   - cuda_compiler_version
-  - cdt_name
   - docker_image
 zlib:
 - '1.2'
--- a/.ci_support/linux_aarch64_c_compiler_version11cuda_compilernvcccuda_compiler_version11.8cxx_compiler_version11fortran_compiler_version11mpi_typeexternal.yaml
+++ b/.ci_support/linux_aarch64_c_compiler_version11cuda_compilernvcccuda_compiler_version11.8cxx_compiler_version11fortran_compiler_version11mpi_typeexternal.yaml
@@ -40,9 +40,9 @@ zip_keys:
 - - c_compiler_version
   - cxx_compiler_version
   - fortran_compiler_version
+  - cdt_name
   - cuda_compiler
   - cuda_compiler_version
-  - cdt_name
   - docker_image
 zlib:
 - '1.2'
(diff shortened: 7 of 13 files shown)


leofang:

leofang:
@dalcinl do you think any of the changes that we made could lead to this file being generated?
```
/home/conda/feedstock_root/build_artifacts/openmpi-mpi_1707129643574/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold/lib/\x01\udce4\x05
```
which seems an odd one to me

dalcinl:
> @dalcinl do you think any of the changes that we made could lead to this file being generated?

I don't think so, but I'll double check our PR diff.

hmaarrfk:
it might be a difference in the lief version... maybe compare and try to pin?

leofang:
liblief/py-lief are both staying on 0.12.3, nothing has changed 🤔

hmaarrfk:
sometimes i download both logs, delete everything until `Z ` and just run `vimdiff` to visually inspect the differences

jakirkham:
Looks like some builds are running out of space. Should we try adding to `conda-forge.yml` (and re-render)?

```yaml
azure:
  free_disk_space: true
```

leofang:

leofang:
No luck, out of disk space doesn't seem to be the issue of having a filename with weird unicode

jakirkham:
Previously saw this error in [this CI log]( https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=871094&view=logs&jobId=517fe804-fa30-5dc2-1413-330699242c05&j=517fe804-fa30-5dc2-1413-330699242c05&t=c7d0fcb4-3a10-5b26-dfea-e879a93fc2ab )

```
conda.CondaMultiError: Error with archive /home/conda/feedstock_root/build_artifacts/pkg_cache/libstdcxx-devel_linux-64-11.4.0-h922705a_105.conda.  You probably need to delete and re-download or re-create this file.  Message was:

failed with error: [Errno 28] No space left on device
Error with archive /home/conda/feedstock_root/build_artifacts/pkg_cache/gfortran_impl_linux-aarch64-11.4.0-hfbda5c0_5.conda.  You probably need to delete and re-download or re-create this file.  Message was:

failed with error: [Errno 28] No space left on device
Error with archive /home/conda/feedstock_root/build_artifacts/pkg_cache/gxx_impl_linux-aarch64-11.4.0-he533754_5.conda.  You probably need to delete and re-download or re-create this file.  Message was:

failed with error: [Errno 28] No space left on device
Error with archive /home/conda/feedstock_root/build_artifacts/pkg_cache/libgfortran5-13.2.0-ha4646dd_5.conda.  You probably need to delete and re-download or re-create this file.  Message was:

failed with error: [Errno 28] No space left on device
Error with archive /home/conda/feedstock_root/build_artifacts/pkg_cache/binutils_impl_linux-64-2.40-hf600244_0.conda.  You probably need to delete and re-download or re-create this file.  Message was:

failed with error: [Errno 28] No space left on device
Error with archive /home/conda/feedstock_root/build_artifacts/pkg_cache/libsanitizer-11.4.0-h4dcbe23_5.conda.  You probably need to delete and re-download or re-create this file.  Message was:

failed with error: [Errno 28] No space left on device
Error with archive /home/conda/feedstock_root/build_artifacts/pkg_cache/gcc_impl_linux-aarch64-11.4.0-he533754_5.conda.  You probably need to delete and re-download or re-create this file.  Message was:

failed with error: [Errno 28] No space left on device
```

leofang:
Yeah we were not concerned with that, we've been focusing on fixing this issue: https://github.com/conda-forge/openmpi-feedstock/pull/142#issuecomment-1927240287, starting here: https://github.com/conda-forge/openmpi-feedstock/pull/141#issuecomment-1915451557.

conda-forge.yml
```diff
@@ -13,3 +13,5 @@ provider:
   linux_ppc64le: default
   osx_arm64: default
 test: native_and_emulated
+azure:
+  free_disk_space: true
```
jakirkham, line 17: Yeah thought there was a chance the two were related. Garbled output like that sometimes shows up when memory usage is exceeded

We could try a pagefile as well


```suggestion
  free_disk_space: true
  settings_linux:
    swapfile_size: 10GiB
```

leofang: Interesting, let's give it a shot. We ran out of tricks anyway...

leofang:

jakirkham:
If that doesn't work, would suggest creating a diff of the dependencies installed when it was working and after it broke. That may shed some light on other relevant changes

jakirkham:
Maybe we could try switching to `patchelf` instead of `LIEF` (for [example]( https://github.com/conda-forge/ctng-compilers-feedstock/blob/5f4387542975181d20803946cb4d2c2061835ccc/recipe/meta.yaml#L501 )):

```yaml
build:
  ...
  rpaths_patcher: patchelf
```

minrk:
[here's a diff of the build logs](https://gist.github.com/minrk/8a82c297e67956b602427cbaa4fbfaba).

the only package difference between bad/good is in the host environment:

```diff
-    openssl:         3.2.1-hd590300_0            conda-forge
-    rdma-core:       50.0-hd3aeb46_0             conda-forge
+    openssl:         3.2.0-hd590300_1            conda-forge
+    rdma-core:       49.0-hd3aeb46_2             conda-forge
```

So first thing to try is probably to pin `rdma-core` to 49 (trying in this PR now)

while the root environment has small differences unlikely to affect things:

```diff
@@ -31,3 +31,3 @@
 conda-index               0.3.0              pyhd8ed1ab_1    conda-forge
-conda-libmamba-solver     24.1.0             pyhd8ed1ab_0    conda-forge
+conda-libmamba-solver     23.12.0            pyhd8ed1ab_0    conda-forge
 conda-oci-mirror          0.1.0              pyhd8ed1ab_0    conda-forge
@@ -103,3 +103,3 @@
 openjpeg                  2.5.0                h488ebb8_3    conda-forge
-openssl                   3.2.1                hd590300_0    conda-forge
+openssl                   3.2.0                hd590300_1    conda-forge
 oras-py                   0.1.14             pyhd8ed1ab_0    conda-forge
@@ -132,3 +132,3 @@
 python_abi                3.10                    4_cp310    conda-forge
-pytz                      2023.4             pyhd8ed1ab_0    conda-forge
+pytz                      2023.3.post1       pyhd8ed1ab_0    conda-forge
 pyyaml                    6.0.1           py310h2372a71_1    conda-forge
```

Other differences:

- kernel version bumped from 6.2.0-1018-azure to 6.2.0-1019-azure
- failed build log has several entries of `ldconfig: $BUILD_DIR/_h_env/lib/ is not a symbolic link` not present in the successful build log

I don't really understand what could be causing that

minrk:
demoting rdma-core didn't fix it. The LIEF check is [unconditional if LIEF is available](https://github.com/conda/conda-build/blob/00a9faff626e16d91fcbc0042011c57585e2e6a9/conda_build/post.py#L605-L608), so selecting patchelf doesn't prevent the use of LIEF. The only way to prevent the failing call is to _remove_ LIEF.

Is there a way to make LIEF unavailable in the conda-build environment on the conda-forge builds? That would allow us to actually just use patchelf.

These two PRs would (each, separately) fix the issue, I think:

- https://github.com/conda/conda-build/pull/5176 (allows LIEF to fail, just like patchelf)
- https://github.com/conda/conda-build/pull/5175 (fixes the encoding error in the warning message, which would prevent the failure)

mbargull:
The `rdma-core=50` and `rdma-core=49.1` builds are faulty in that `ldconfig -vn "${PREFIX}/lib"` creates symlinks for `libmana`/`libmlx5` with those broken filenames.
The `SONAME`s of the libraries look alright, though.
IDK, why it happens but I'm currently trying to replicate it.

mbargull:
Seems to be an issue with `conda-build>=3.28`; still investigating...

mbargull:
> Seems to be an issue with `conda-build>=3.28`; still investigating...

There is a small-ish bug in `conda-build>3.28` which lets it run `patchelf` not only for the actual binary but also its symlinks.
In this case we have symlinks like `libibverbs/libmana-rdmav34.so -> ../libmana.so.1.0.50.0` in `rdma-core` which, when used as `patchelf`'s import, sets the rpath to for `libmana.so.1.0.50.0` to `$ORIGIN/.:$ORIGIN/..` instead of `$ORIGIN/.`.
This is of course wrong, but shouldn't cause too much problems.
Unfortunately, we then run into a `patchelf` bug which leads to `ldconfig` (called in downstream build/install scripts like here) creating additional symlinks with non-UTF-8 filenames.
I'll write an issue, test and fix for `conda-build` and link it here later.

leofang:

leofang:
@conda-forge-admin, please restart ci

leofang:
It seems @mbargull's PR https://github.com/conda/conda-build/pull/5181 was merged and available in conda-build 24.1.2, but we need to bump the pinned version in conda-forge-ci-setup to use it.

h-vetinari:
> but we need to bump the pinned version in conda-forge-ci-setup to use it.

seems we need to remove [this line](https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/e403f01fd447e48ba3a777a9b642354ab1b7e746/recipe/meta.yaml#L72) now that conda-build uses CalVer - care to send a PR?

leofang:
> seems we need to remove [this line](https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/e403f01fd447e48ba3a777a9b642354ab1b7e746/recipe/meta.yaml#L72) now that conda-build uses CalVer - care to send a PR?

Sure, see https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/304.

leofang:
@conda-forge-admin, please restart ci

leofang:
> > seems we need to remove [this line](https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/e403f01fd447e48ba3a777a9b642354ab1b7e746/recipe/meta.yaml#L72) now that conda-build uses CalVer - care to send a PR?
> 
> Sure, see [conda-forge/conda-forge-ci-setup-feedstock#304](https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/304).

It seems conda-build is still pinned at 3.28.4. @h-vetinari @beckermr any idea where did I miss to modify?

leofang:
I guess the pinned `conda-build` might come from the docker images. Can any of you rebuild the images with the latest conda-build? It seems possible after @jakirkham's PR: https://github.com/conda-forge/docker-images/pull/230

h-vetinari:
Have a look at this:
https://github.com/conda-forge/openmpi-feedstock/blob/efdc204c50e34b01b95170d54debd6d97def2d1f/.scripts/build_steps.sh#L36-L39

Modifying this to require newest conda-build gives a long resolution error:
```
>mamba create -n test pip mamba conda-build=24 boa conda-forge-ci-setup=4
```
Paring this down a bit by pinning to the last boa version (that should by all appearances [not be pinned](https://github.com/conda-forge/boa-feedstock/blob/02c8053078c2a650c7063d2d8e1b33bd09ee73c4/recipe/meta.yaml#L26)) yields:
```
>mamba create -n test pip mamba conda-build=24 boa=0.16 conda-forge-ci-setup=4

Looking for: ['pip', 'mamba', 'conda-build=24', 'boa=0.16', 'conda-forge-ci-setup=4']

conda-forge/win-64                                          Using cache
conda-forge/noarch                                          Using cache
Could not solve for environment specs
The following packages are incompatible
├─ boa 0.16**  is installable and it requires
│  └─ conda-build >=3.24,<24.1.0a0 , which can be installed;
└─ conda-build 24**  is not installable because it conflicts with any installable versions previously reported.
```

Thus we find https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/657 due to https://github.com/mamba-org/boa/issues/392. There's discussion about unblocking conda-build 24 in https://github.com/conda-forge/conda-smithy/pull/1844 already, and I just opened https://github.com/conda-forge/boa-feedstock/issues/79 for an alternative approach that's slightly less intrusive (IMO).

jakirkham:
Wouldn't we also need to rebuild `rdma-core` first to produce packages with the fix for `openmpi` to use?

Admittedly I've not been following this as closely as others, so that could be totally off base

xhochy:
@conda-forge-admin please rerender

leofang:
@mbargull @h-vetinari any other thoughts? 😛 Now we are using the latest conda-build, but we still hit the same error...

jakirkham:
Am curious what others think of my question above: https://github.com/conda-forge/openmpi-feedstock/pull/142#issuecomment-1955964849

Suspect that is an important next step

leofang:
> Am curious what others think of my question above: https://github.com/conda-forge/openmpi-feedstock/pull/142#issuecomment-1955964849

Based on @mbargull's [earlier analysis](https://github.com/conda-forge/openmpi-feedstock/pull/142#issuecomment-1938222548), the CI failed because we had two unrelated bugs acting jointly:
1. conda-build
2. patchelf (calling `ldconfig` that created the buggy filenames, see also [here](https://github.com/conda-forge/openmpi-feedstock/pull/142#issuecomment-1936503202))

I was hoping that with all the fixes we now eliminate Bug 1, thereby stopping the joint action of both bugs, as we are supposed to only have Bug 2 left and it should be relatively harmless. If my understanding is correct, then it means we still have a gap in this analysis.

h-vetinari:
> Wouldn't we also need to rebuild `rdma-core` first to produce packages with the fix for `openmpi` to use?

AFAIU, as long as that malformed (non-UTF-8) path is there - which is still the case - the rdma rebuild shouldn't make a difference. Unless we completely get rid of the symlinks in that package, but that's not a reasonable ask; we should IMO fix (or work around) the bug in patchelf/conda-build.

leofang:
@mbargull any change you have further insight on this issue? 🙂

mbargull:
> @mbargull any change you have further insight on this issue? 🙂

Sorry, didn't follow this issue; looks like `rdma-core` has not been rebuilt yet.
The libraries therein are broken for the versions built with `conda-build >=3.28,<24.1.2`, see my previous comment https://github.com/conda-forge/openmpi-feedstock/pull/142#issuecomment-1936503202 :
> The `rdma-core=50` and `rdma-core=49.1` builds are faulty in that `ldconfig -vn "${PREFIX}/lib"` creates symlinks for `libmana`/`libmlx5` with those broken filenames.

leofang:
So was my understanding in https://github.com/conda-forge/openmpi-feedstock/pull/142#issuecomment-1965662764 incorrect, that there are still issues to fix, regardless whether we fixed conda-build (1)?

mbargull:
> So was my understanding in [#142 (comment)](https://github.com/conda-forge/openmpi-feedstock/pull/142#issuecomment-1965662764) incorrect, that there are still issues to fix, regardless whether we fixed conda-build (1)?

Yes, it is partially incorrect. `patchelf` does not call `ldconfig`.
There's no "gap in this analysis"; it's simply that the combination of bugs in `conda-build` and `patchelf` lead to faulty artifacts (in `rdma-core`) which when processed downstream (this feedstock) with `ldconfig` causes the creation of symlinks with broken names.

recipe/meta.yaml
```diff
@@ -51,6 +52,7 @@ outputs:
         - libevent
         - libnl  # [linux]
         - zlib
+        - rdma-core 49  # [linux]
```
leofang, line 55: We should just let it float. This was added for debugging purpose.
```suggestion
```

leofang:

recipe/meta.yaml
```diff
@@ -22,6 +22,7 @@ build:
   number: {{ build }}
   skip: true  # [win]
   skip: true  # [linux and cuda_compiler_version != "11.8"]
+  rpaths_patcher: patchelf  # [linux]
```
leofang, line 25: This should be unneeded, let's remove it once we confirm rebuilding rdma-core was the last blocker
```suggestion
```

leofang:
> Yes, it is partially incorrect. `patchelf` does not call `ldconfig`. (...) which when processed downstream (this feedstock) with `ldconfig` causes the creation of symlinks with broken names.

Forgive my ignorance, so when exactly is `ldconfig` called and by whom?

mbargull:
> @mbargull Forgive my ignorance, so when exactly is `ldconfig` called and by whom?

https://www.gnu.org/software/libtool/manual/libtool.html#Finish-mode
(You should find the exact invocations in a `configure` scripts, but build's source archive might also carry a `libtool.m4`  which would be slightly more readable.)

leofang:
linux-64 CI is finally happy!

leofang:

leofang (approved):

leofang:
Thanks @mbargull @minrk @jakirkham and everyone for helping out! This is one of the most tricky bugs I've ever seen on conda-forge!

Issue #153, opened 2024-04-22
OpenMPI dependency resolving to `external` build


folmos-at-orange:
### Comment:

I'm building a package depending on `openmpi`. I pinned the run dependency version to `4.1.6` (which is on conda-forge). However when installing the package it sometimes resolve to the `external` build. As a consequence when I install the package it doesn't work because the external build doesn't install any binaries (as expected).

Any idea on why this happens?

I might want to pin to `openmpi=4.1.6=h*` but that would defeat the purpose of having external builds.

Thanks in advance

dalcinl:
I'm wondering whether these packages are somehow broken. I would recommend you contact the conda-forge team via chat, I don't know how to help you further. Perhaps making a new build of that version would help, but I'm not sure.

leofang:
I am clueless too, as we did ensure the non-external builds have a higher priority 
https://github.com/conda-forge/openmpi-feedstock/blob/16df999278a105ae143deacc96680b799d79602a/recipe/meta.yaml#L5-L8
maybe you can share the outcome of `conda list` and examine if there are potential conflicts?

folmos-at-orange:
I really don't understand what is happening. Repoquery resolve its ok, but install still installs the `external` build. Below some outputs for the a version of my package with only `openmpi=4.1.4` pinned.

**conda info**
```
conda info
    active environment : test
    active env location : /home/runner/miniconda3/envs/test
            shell level : 2
       user config file : /home/runner/.condarc
 populated config files : /home/runner/.condarc
          conda version : 24.3.0
    conda-build version : not installed
         python version : 3.12.2.final.0
                 solver : libmamba (default)
       virtual packages : __archspec=1=zen2
                          __conda=24.3.0=0
                          __glibc=2.31=0
                          __linux=5.15.0=0
                          __unix=0=0
       base environment : /home/runner/miniconda3  (writable)
      conda av data dir : /home/runner/miniconda3/etc/conda
  conda av metadata url : None
           channel URLs : https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/linux-64
                          https://repo.anaconda.com/pkgs/r/noarch
          package cache : /home/runner/conda_pkgs_dir
       envs directories : /home/runner/miniconda3/envs
                          /home/runner/.conda/envs
               platform : linux-64
[...]

PR #156, opened 2024-05-29
ENH ensure external variants have same deps as regular ones for the solver

beckermr:
Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)

xref #153 

I am hoping this will help with #153 above in that the external builds will be identical expect that they are shells, have an extra feature, and a lower build number. This should help the solver deprioritize them properly. If this works, we'd need to possibly mark some of the older 5.x ones broken.

### DIFF
--- a/.ci_support/migrations/libhwloc2100.yaml
+++ b/.ci_support/migrations/libhwloc2100.yaml
@@ -1,8 +1,0 @@
-__migrator:
-  build_number: 1
-  commit_message: Rebuild for libhwloc 2.10.0
-  kind: version
-  migration_number: 1
-libhwloc:
-- 2.10.0
-migrator_ts: 1711507406.8423882
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,7 +1,7 @@
 {% set version = "5.0.3" %}
 {% set major = version.rpartition('.')[0] %}
 {% set cuda_major = (cuda_compiler_version|default("11.8")).rpartition('.')[0] %}
-{% set build = 2 %}
+{% set build = 3 %}
 
 # give conda package a higher build number
 {% if mpi_type == 'conda' %}
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -24,12 +24,19 @@ build:
   skip: true  # [linux and cuda_compiler_version != "11.8"]
 
 outputs:
-  {% if mpi_type == 'conda' %}
   - name: openmpi
+    {% if mpi_type == 'conda' %}
     script: build-mpi.sh
+    {% endif %}
     build:
+      {% if mpi_type == 'conda' %}
       run_exports:
         - {{ pin_subpackage('openmpi', min_pin='x.x.x', max_pin='x') }}
+      {% else %}
+      string: {{ mpi_type }}_h{{ PKG_HASH }}_{{ build }}
+      track_features:
+        - openmpi_{{ mpi_type }}
+      {% endif %}
       ignore_run_exports:
         - ucx  # [linux]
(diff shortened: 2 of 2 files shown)


beckermr:
@conda-forge-admin rerender

beckermr:
@conda-forge/openmpi I propose we merge this one and try it out.

jaimergp:
It could work but the older leaner versions will still be around so I wonder if the solver will still prefer those?

minrk:
IIUC, this means that unused dependencies get added to the environment when you install openmpi-external, to ensure that any given external build has all the same dependency conflicts/compatibilities with other packages as the 'real' builds?

I'm okay with this for now if it solves a problem affecting users, but it doesn't quite seem like the right thing to do in the long run. openmpi-external shouldn't really be considered for installation without _matching_ openmpi on the host.

 Is there any chance we could come up with a way to add virtual packages to conda itself, so openmpi-external could require e.g. `__openmpi=5.0.1`?

It seems to me like openmpi-external should really be on a different _channel_ (e.g. `conda-forge/label/openmpi-external`) to make it truly opt-in instead of deprioritizing by build number, which lets the solver consider it a valid candidate to conflict resolution, which it never is.

jaimergp:
> Is there any chance we could come up with a way to add virtual packages to conda itself, so openmpi-external could require e.g. __openmpi=5.0.1?

There's a virtual plugin hook we can use now, but users would need to install it next to `conda` in `base`

> It seems to me like openmpi-external should really be on a different channel (e.g. conda-forge/label/openmpi-external) to make it truly opt-in instead of deprioritizing by build number, which lets the solver consider it a valid candidate to conflict resolution, which it never is.

I like this!

beckermr:
I'm very much against no a different channel. The solver should be able to handle this fine and the old conda solver did for a long time.

beckermr:
Just to expand my comments a bit, the correct solver behavior is to globally minimize the number of track_features in the env. The mamba solver explicitly does not do this. Most of the time, this produces a correct solution. However, in this case it does not. 

So, the issue here lies with the solver, not the fact that we've somehow mixed packages on a single channel that should really be on separate channels.

beckermr:
> I'm okay with this for now if it solves a problem affecting users, but it doesn't quite seem like the right thing to do in the long run. openmpi-external shouldn't really be considered for installation without matching openmpi on the host.

Yes. This is a correct solution and would make things easier for folks. We can probably get by with a global conda-forge-conda-plugins package that holds this virtual package, one for mpich, and eventually anything else we need in the future.

minrk:
I'm fine with whatever the simplest practical solution is. I generally don't trust relying on _priority_ to enforce something that should _never_ happen implicitly, since it so often results in failures like #153.

I do find the virtual package approach appealing, if it's doable.

👍 to merging this as-is for now and marking all previous 5.0 external builds as broken, since it should help in the short term.

beckermr:
Sounds good @minrk. I'll take care of marking the other 5 external builds as broken and we'll see if this helps. @jaimergp is also looking into improving things on the mamba side.

jaimergp:
> I'm happy to start a repo and help build this out. I'll need to read up on how to write plugins for conda. :)

I can provide the boilerplate if you want! They also have a template repo somewhere in the conda org.

beckermr:
Awesome. We should do some investigation on how to detect the openmpi version as well.

beckermr:
This works

```
ompi_info --parsable | grep "ompi:version:full:" | cut -d ":" -f 4
```

PR #157, opened 2024-06-03
set environment in activate.sh for conda-build

minrk:
this should avoid the tedious requirement to set a bunch of ever-changing environment variables in every package that builds against openmpi

I did not copy allow_root, which appears to no longer be necessary with latest conda-forge infrastructure

### DIFF
--- a/recipe/run_test.sh
+++ b/recipe/run_test.sh
@@ -1,12 +1,6 @@
 #!/bin/bash
 set -ex
 
-export OMPI_MCA_pml=ob1
-export OMPI_MCA_btl=sm,self
-export OMPI_MCA_plm_ssh_agent=false
-export OMPI_MCA_rmaps_default_mapping_policy=:oversubscribe
-export OMPI_ALLOW_RUN_AS_ROOT=1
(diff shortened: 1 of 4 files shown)


beckermr (approved):

recipe/run_test.sh
```diff
@@ -1,12 +1,6 @@
 #!/bin/bash
 set -ex
 
-export OMPI_MCA_pml=ob1
-export OMPI_MCA_btl=sm,self
-export OMPI_MCA_plm_ssh_agent=false
-export OMPI_MCA_rmaps_default_mapping_policy=:oversubscribe
-export OMPI_ALLOW_RUN_AS_ROOT=1
```
dalcinl, line 8: @minrk Are you absolutely sure `mpiexec` will never be invoked by the root user?

minrk: I don't believe any commands inside recipes are invoked as root in conda-forge at least, which is my main goal to cover here. Additional builds elsewhere are still free to specify additional arguments if they need them, as they have always needed to do before this.

dalcinl (approved):

PR #158, opened 2024-06-05
add wrapper compiler flags in CONDA_BUILD, install .mod in include

minrk:
openmpi appears to be the only fortran package that puts .mod files in $PREFIX/lib, which is not found by default, so move them to include like everybody else, so gfortran finds them with standard include flags.

Without this, some situations (only found so far in mac cross-compile) need `-I$PREFIX/lib` to find fortran modules.

also ensures that OMPI_CC=$CC and friends are set in conda-build, which is generally necessary for cross-compilation and should now be more likely to work by default.

found via https://github.com/conda-forge/scalapack-feedstock/pull/34

### DIFF
--- a/recipe/build-mpi.sh
+++ b/recipe/build-mpi.sh
@@ -73,6 +73,10 @@ fi
 make -j"${CPU_COUNT:-1}"
 make install
 
+# openmpi installs .mod files in the wrong prefix (/lib instead of /include)
+ls -l $PREFIX/lib/*.mod
+mv -v $PREFIX/lib/*.mod $PREFIX/include/
+
 POST_LINK=$PREFIX/bin/.openmpi-post-link.sh
 if [ -n "$build_with_ucx" ]; then
     echo "setting MCA pml to ^ucx..."
--- a/recipe/build-mpi.sh
+++ b/recipe/build-mpi.sh
@@ -74,8 +74,11 @@ make -j"${CPU_COUNT:-1}"
 make install
 
 # openmpi installs .mod files in the wrong prefix (/lib instead of /include)
-ls -l $PREFIX/lib/*.mod
-mv -v $PREFIX/lib/*.mod $PREFIX/include/
+# symlink instead of copy to avoid breaking anything (unlikely)
+for f in $PREFIX/lib/*.mod; do
+  modname=$(basename "$f")
+  ln -sv "../lib/${modname}" "$PREFIX/include/${modname}"
+done
 
 POST_LINK=$PREFIX/bin/.openmpi-post-link.sh
 if [ -n "$build_with_ucx" ]; then
--- a/recipe/openmpi_activate.sh
+++ b/recipe/openmpi_activate.sh
@@ -6,6 +6,7 @@ if [[ "${CONDA_BUILD:-}" = "1" ]]; then
       declare -x "OMPI_${_var}=${!_var}"
     fi
   done
+  export OPAL_PREFIX="$PREFIX"
 
   # runtime variables
   export OMPI_MCA_plm_ssh_agent=false
--- a/recipe/openmpi_activate.sh
+++ b/recipe/openmpi_activate.sh
@@ -7,7 +7,8 @@ if [[ "${CONDA_BUILD:-}" = "1" ]]; then
       export OMPI_${_var}="${!_var}"
     fi
     if [[ -z "${OMPI_FCFLAGS}" && ! -z "${FFLAGS}" ]]; then
-      export OMPI_FCFLAGS="$FFLAGS"
+      echo OMPI_FCFLAGS="-I$PREFIX/include $FFLAGS"
+      export OMPI_FCFLAGS="-I$PREFIX/include $FFLAGS"
     fi
   done
   export OPAL_PREFIX="$PREFIX"
--- a/recipe/openmpi_activate.sh
+++ b/recipe/openmpi_activate.sh
@@ -4,7 +4,7 @@ if [[ "${CONDA_BUILD:-}" = "1" ]]; then
   for _var in CC CXX FC CPPFLAGS CFLAGS CXXFLAGS FCFLAGS LDFLAGS; do
     if [[ ! -z "${!_var:-}" ]]; then
       echo "OMPI_${_var}=${!_var}"
-      declare -gx "OMPI_${_var}=${!_var}"
+      export OMPI_${_var}="${!_var}"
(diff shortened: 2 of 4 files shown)


recipe/run_test.sh
```diff
@@ -26,44 +26,58 @@ if [[ $PKG_NAME == "openmpi" ]]; then
   command -v mpiexec
   $MPIEXEC --help
   $MPIEXEC -n 4 ./helloworld.sh
+
+  test -f $PREFIX/include/mpi.mod
+  test ! -f $PREFIX/include/mpi.mod
```
beckermr, line 31: Shouldn't this be PREFIX/lib?

minrk: yes, fixed.

beckermr (approved):

dalcinl:
@minrk What about `LDFLAGS` ?

recipe/build-mpi.sh
```diff
@@ -73,6 +73,10 @@ fi
 make -j"${CPU_COUNT:-1}"
 make install
 
+# openmpi installs .mod files in the wrong prefix (/lib instead of /include)
+ls -l $PREFIX/lib/*.mod
+mv -v $PREFIX/lib/*.mod $PREFIX/include/
```
dalcinl, line 78: What about fixing the include  paths?

```
$ grep 'compiler_flags=' share/openmpi/mpif*-wrapper-data.txt 
share/openmpi/mpif77-wrapper-data.txt:compiler_flags=-I${includedir}   -I${libdir}
share/openmpi/mpif90-wrapper-data.txt:compiler_flags=-I${includedir}   -I${libdir}
share/openmpi/mpifort-wrapper-data.txt:compiler_flags=-I${includedir}   -I${libdir}
```

dalcinl: Using `mv` could this have backward compatibility implications. Maybe you should keep symlinks in `libdir` ?

minrk: I think it's quite unlikely, since mpifort puts both $PREFIX/include and $PREFIX/lib on the include path (with OMPI_FCFLAGS, it's only $PREFIX/include) but we can copy to be sure.

minrk: used symlinks

dalcinl: I was thinking about users setting flags "by hand", that is, not using MPI compiler wrappers.

dalcinl (approved):

recipe/build-mpi.sh
minrk: I don't think they can, because all the `.h` headers like `mpif.h` are in include, so openmpi needs _both_ `-I$PREFIX/include -I$PREFIX/lib`. I'm not sure if there are any feasible situations where:

1. `mpifort` isn't used (not affected)
2. `-I$PREFIX/lib` is given, and
3. `-I$PREFIX/include is _not_ given (i.e. no 'normal' headers or modules are required)

all of which must be true for moving to break anything.

dalcinl: Those using `use mpi` or `use mpi_f08` modules do not need `mpif.h` but just the `mpi_*.mod` files. Therefore it could be possible that some folks where using  a bare `gfortran -I$PREFIX/lib -L$PREFIX/lib ...` and things would just work, but then things would break after your changes if you had not kept symlinks in libdir.

minrk: Yeah, I guess it's technically not impossible, though they would have to use mpi, nothing else, and only openmpi.

msarahan:
I hate to be "that guy" who comes in and says "this broke me," but I'm even worse: "that guy, who's coming in on behalf of others, saying "this broke them."" Cugraph is having to pin build numbers to avoid this change: https://github.com/rapidsai/cugraph/issues/4474

It looks to me like this change affects CMake somehow, which is what breaks cugraph. Does it make sense to try to add a test in this recipe that has a simple CMakeLists.txt project that calls FindMPI from an env with this package in it? 

We can contribute this if you'd like, but I admit I'm not super familiar here and it make take me some time.

leofang:
Contribution is welcome! Although it's unclear yet to me what's causing the issue. By a quick glimpse of FindMPI, it seems it's indeed expected that .mod files should be found in the include dir:
https://github.com/Kitware/CMake/blob/48827cd746895da872e3f5834ed749439e83e77e/Modules/FindMPI.cmake#L1184-L1189
so it could be something else.

leofang:
@minrk i noticed the new build activate script (that's supposed to be run only in conda-forge CI) is also now run in RAPIDS CI (because they use mambabuild). Can we tighten the check condition to check more than CONDA_BUILD? Maybe it's causing issues.
https://github.com/rapidsai/cugraph/actions/runs/9402241384/job/25895993301#step:7:721

recipe/openmpi_activate.sh
```diff
@@ -1,5 +1,19 @@
...
+    if [[ ! -z "${!_var:-}" ]]; then
+      echo "OMPI_${_var}=${!_var}"
+      export OMPI_${_var}="${!_var}"
+    fi
+    if [[ -z "${OMPI_FCFLAGS}" && ! -z "${FFLAGS}" ]]; then
+      echo OMPI_FCFLAGS="-I$PREFIX/include $FFLAGS"
+      export OMPI_FCFLAGS="-I$PREFIX/include $FFLAGS"
+    fi
+  done
+  export OPAL_PREFIX="$PREFIX"
```
leofang, line 14: Maybe these are causing issues?

minrk:
No apologies necessary, sorry for causing trouble and thanks for reporting!

If you can share debug output, that would help. I can't actually find my way to what's actually failing with all the indirection and abstraction in that CI setup, so I don't know what's causing it to fail.

Are any of the failing builds cross compiling?

I think passing through the compiler flags may have been the wrong choice, and puts way more in the base compiler wrapper command than is there by default. Perhaps we should roll those back and leave just the compiler envs themselves and OPAL_PREFIX, which I think are very unlikely to cause a problem, are sensible defaults, and easily overridden (or unset). None of those are conda-forge specific, I think.

Is it possible to test if unsetting :

```
unset OMPI_CPPFLAGS
unset OMPI_CFLAGS
unset OMPI_CXXFLAGS
unset OMPI_FCFLAGS
unset OMPI_LDFLAGS
```

fixes it?

I think perhaps the compiler env should only be set for cross-compilation, where we know the default values are wrong. THat's what #159 does.

Issue #143, opened 2024-02-06
mpifort 5.0.1 fails with undefined refs to memcpy@GLIBC_2.14 and clock_gettime@GLIBC_2.17


charlesgwaldman:
### Solution to issue cannot be found in the documentation.

- [X] I checked the documentation.

### Issue

I am using cmake, gfortran, and openmpi from conda-forge to compile a Fortran package.   With cmake 3.28.3, gfortran 13.2.0 and openmpi 5.0.0 everything works.  When openmpi upgraded to the latest 5.0.1 version I started getting this error:
```
-- Check for working Fortran compiler: /home/cgw/miniforge3/envs/mfix-git/bin/mpifort
-- Check for working Fortran compiler: /home/cgw/miniforge3/envs/mfix-git/bin/mpifort - broken
CMake Error at /home/cgw/miniforge3/envs/mfix-git/share/cmake-3.28/Modules/CMakeTestFortranCompiler.cmake:59 (message):
  The Fortran compiler

    "/home/cgw/miniforge3/envs/mfix-git/bin/mpifort"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: '/tmp/TTT/build_dmp/CMakeFiles/CMakeScratch/TryCompile-38w6ll'
    
    Run Build Command(s): /home/cgw/miniforge3/envs/mfix-git/bin/cmake -E env VERBOSE=1 /usr/bin/gmake -f Makefile cmTC_6144f/fast
    /usr/bin/gmake  -f CMakeFiles/cmTC_6144f.dir/build.make CMakeFiles/cmTC_6144f.dir/build
    gmake[1]: Entering directory '/tmp/TTT/build_dmp/CMakeFiles/CMakeScratch/TryCompile-38w6ll'
    Building Fortran object CMakeFiles/cmTC_6144f.dir/testFortranCompiler.f.o
    /home/cgw/miniforge3/envs/mfix-git/bin/mpifort    -c /tmp/TTT/build_dmp/CMakeFiles/CMakeScratch/TryCompile-38w6ll/testFortranCompiler.f -o CMakeFiles/cmTC_6144f.dir/testFortranCompiler.f.o
    Linking Fortran executable cmTC_6144f
    /home/cgw/miniforge3/envs/mfix-git/bin/cmake -E cmake_link_script CMakeFiles/cmTC_6144f.dir/link.txt --verbose=1
[...]

PR #159, opened 2024-06-08
only set wrapper args for cross compilation, keeping subset of LDFLAGS necessary for sysroot < 2.17

minrk:
defaults are not likely to be wrong for native compilation, don't mess with them, I think #158 went a bit too far.

and don't passthrough $LDFLAGS, etc. from compiler activation, set the actual correct values from the wrappers with the right prefix.

closes #163

### DIFF
--- a/conda-forge.yml
+++ b/conda-forge.yml
@@ -12,6 +12,8 @@ conda_forge_output_validation: true
 github:
   branch_name: main
   tooling_branch_name: main
+os_version:
+  linux_64: cos7
 provider:
   linux_aarch64: default
   linux_ppc64le: default
--- a/conda-forge.yml
+++ b/conda-forge.yml
@@ -14,6 +14,8 @@ github:
   tooling_branch_name: main
 os_version:
   linux_64: cos7
+  linux_aarch64: cos7
+  linux_ppc64le: cos7
 provider:
   linux_aarch64: default
   linux_ppc64le: default
--- a/recipe/build-mpi.sh
+++ b/recipe/build-mpi.sh
@@ -11,6 +11,7 @@ export FC=$(basename "$FC")
 unset FFLAGS F77 F90 F95
 
 # tweak compiler flags
+wrapper_ldflags="-L$PREFIX/lib -Wl,-rpath,$PREFIX/lib"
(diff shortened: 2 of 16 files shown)


minrk:
Still need to update the tests

recipe/build-mpi.sh
```diff
@@ -23,6 +24,9 @@ fi
 build_with_ucx=""
 if [[ "$target_platform" == linux-* ]]; then
     build_with_ucx="--with-ucx=$PREFIX"
+    # allow-shlib-undefined required for dependencies to link against older sysroot
+    # avoids undefined
+    wrapper_ldflags="${wrapper_ldflags} -Wl,--allow-shlib-undefined"
```
minrk, line 29: This is actually the change needed to keep downstream packages from running into the glibc version, as verified by the fact that the tests still pass and still run with 3.12, and do not pass the conda-forge $LDFLAGS.

This change should actually fix #143

conda-forge.yml
```diff
@@ -12,6 +12,8 @@ conda_forge_output_validation: true
 github:
   branch_name: main
   tooling_branch_name: main
+os_version:
+  linux_64: cos7
```
jakirkham, line 16: Even though this is already the default in the other cases (and soon will be for `linux-64`), maybe we should note these explicitly in all cases?

```suggestion
  linux_64: cos7
  linux_aarch64: cos7
  linux_ppc64le: cos7
```

We could also then drop these lines from `conda_build_config.yaml` as `os_version` also configures `c_stdlib_version` for us

https://github.com/conda-forge/openmpi-feedstock/blob/99059c0e090c0f8b6f821cfd5e80c0b1c35f62f0/recipe/conda_build_config.yaml#L1-L2

minrk:
I'd really like to land this to unbreak things caused by #158. If folks are still uncomfortable with including the link flag from $LDFLAGS, I can remove it and isolate that to its own PR.

minrk:
I will merge this today unless there are objections, to prevent more conflicts like #162, since all questions raised so far have been addressed. To reiterate, this PR applies a strict subset of flags than the currently published build, it does not add any. It's just that one (the one that's necessary for building in conda-forge default environments) is now called out by name.

PR #160, opened 2024-06-09
add sysroot lower bound to run_constrained

minrk:
prevents downstream packages from building with older sysroot (#143)

draft because:

- it's unclear if this is an openmpi-specific issue or should be done in sysroot's run_exports
- may need repodata patches for prior 2.17 builds for solver consistency if this is the right thing to do
- it seems like we need guidance from conda-forge core on how best to signal that "when build with sysroot X, _linking against this package_ must also build with sysroot >=X", and when that's likely to be the case

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,7 +1,7 @@
 {% set version = "5.0.3" %}
 {% set major = version.rpartition('.')[0] %}
 {% set cuda_major = (cuda_compiler_version|default("11.8")).rpartition('.')[0] %}
-{% set build = 5 %}
+{% set build = 6 %}
 
 # give conda package a higher build number
 {% if mpi_type == 'conda' %}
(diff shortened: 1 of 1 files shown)


h-vetinari:
> it's unclear if this is an openmpi-specific issue or should be done in sysroot's run_exports

Given the discussion in https://github.com/conda-forge/linux-sysroot-feedstock/issues/63, I'd suggest you add this to openmpi for now; if at any point it becomes added as a sysroot constraint itself, you can always remove it later.

minrk:
Thanks! That's useful. If I've read that discussion correctly, it won't have the desired effect when `openmpi` is just in `host`, right? That limits the usefulness a bit, I suppose, since that's all linux-64 builds, which are the main ones affected by the current issue.

minrk:
I changed my mind, on re-reading various situations from #143, adding this constraint will make it more likely for compilation to succeed in _user_ environments, since building with compiler wrappers is something that's normal to occur for users at runtime, not just building packages in conda-build.

dalcinl:
@minrk What's the status of this PR? Are you waiting for anything to happen upstream in conda-forge? Or it is just an abandoned idea?

minrk:
I don't think this really makes sense, so I don't think we should do special things for openmpi here.

PR #162, opened 2024-06-12
Make activate script POSIX-compatible.

ktlim:
The activate script can be run in a number of shells, not just bash, so it should only use POSIX-compatible syntax.

Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [x] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #163, opened 2024-06-13
Add more POSIX shell compatibility.

ktlim:
`${!}` indirection is also a bash-only feature.
Also set `OMPI_FCFLAGS` once.

Checklist

### DIFF
--- a/.ci_support/linux_64_c_compiler_version11c_stdlib_version2.17cuda_compilernvcccuda_compiler_version11.8cxx_compiler_version11fortran_compiler_version11mpi_typeconda.yaml
+++ b/.ci_support/linux_64_c_compiler_version11c_stdlib_version2.17cuda_compilernvcccuda_compiler_version11.8cxx_compiler_version11fortran_compiler_version11mpi_typeconda.yaml
@@ -46,4 +46,4 @@ zip_keys:
   - cuda_compiler_version
   - docker_image
 zlib:
-- '1.2'
+- '1'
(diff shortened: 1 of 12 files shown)


ktlim:

minrk:
Thanks for the PR! let's please solve this in #159 instead of separately. Instead of merging this, I'll fix it in #159 to avoid introducing more conflicts.

Can I ask how you encountered this? I've been testing with at least zsh, bash, and sh, all of which don't have an issue. conda-build specifies that build scripts are run in bash, so the `if $CONDA_BUILD` part should only. But maybe some shells do processing of the whole file differently.

minrk (changes requested):
will be applied in #159

ktlim:
We only encountered the problem with the first line (which was why that was the only fix at first) trying to activate `openmpi` on Debian under `/bin/sh` (which is not `bash`).  I thought I had looked around for documentation on conda-build's shell but now I see that it is indeed specified to be bash, so this PR wasn't really necessary.

beckermr:
I guess it depends on how much of the file the shell interprets. If it goes line by line, then we could have not-supported syntax tucked in an if statement and it'd be fine. IDK how the shell actually does things.

minrk:
@ktlim thanks! That's my understanding, and thanks for the fix in #162. We can come back with more fixes if someone encounters a problem, but at least for now I think we can say that this is `sh`-compatible, and I've tested that it works with `/bin/sh` for me, at least.

timj:
It looks like the ticket that closed this made things worse and I see that the fixes from #162 got reverted last month so the script errors out all the time.

```
#42 482.4 ./lsstinstall: 3: /opt/lsst/software/stack/conda/envs/lsst-scipipe-11.0.0/etc/conda/activate.d/openmpi_activate.sh: [[: not found
#42 482.4 ./lsstinstall: 3: /opt/lsst/software/stack/conda/envs/lsst-scipipe-11.0.0/etc/conda/activate.d/openmpi_activate.sh: -n: not found
```

Something is seemingly wrong in terms of whether these .sh have to be posix or not.

Issue #119, opened 2023-03-25
UCX 1.14.0 packages make CUDA optional


jakirkham:
As part of the UCX 1.14.0 upgrade ( https://github.com/conda-forge/ucx-split-feedstock/pull/111 ), CPU & GPU builds were merged making `cudatoolkit` optional and `ucx-proc` unneeded. So it should be possible to use UCX more generally

dalcinl:
I have no idea how these improvements should affect the openmpi package. @leofang ?

jakirkham:
Think it comes down to changing this...

https://github.com/conda-forge/openmpi-feedstock/blob/dfc3d61785630b94cd597292fc0c8e66f1ce3887/recipe/meta.yaml#L47-L48

...to this...

```yaml
 - ucx                 # [enable_cuda]
```

<hr>

Though `ucx` should work on Linux (even without CUDA)

Tried playing with some changes to capture this in PR ( https://github.com/conda-forge/openmpi-feedstock/pull/121 ). Though there may be other valid approaches

Would defer to Leo and you on what makes the most sense here 🙂

dalcinl:
@jakirkham @minrk Please help me decide what to do in #128. 

- [ ] From @minrk's comment, looks like I have to remove the requirement `ucx-proc =*=gpu`.
- [ ] What about the `cudatoolkit  >= {{ cudatoolkit }}` requirement? Should I remove it?

dalcinl:
BTW, I brancked of `4.x`. Should we generate fresh builds with updated requirements?

leofang:
Sorry I completely missed this. Why can't `openmpi` just depend on `ucx` unconditionally? `ucx` can also be used for CPU-only transfers, and based what I learned from @pentschev (IIRC) `ucx` wouldn't do anything special if the CUDA presence is not detected, meaning we can just do this
```yaml
# the actual content would be more convolved due to the need of supporting CUDA 11 & 12,
# as they have different package layout. This is for illustration only.
run_constrained:
  - cudatoolkit 
```
Once we decide how to approach this, I can take care of the CUDA part.

dalcinl:
> Why can't `openmpi` just depend on `ucx` unconditionally?

Maybe we will have to do it anyway. v5.0.x has changed the way they do things by default, components are no longer built as plugins, but rather linked to `libmpi.so` (see discussion in #128)

leofang:
Closing. In #128 we use UCX unconditionally.

[...]

Issue #102, opened 2022-05-05
Latest Linux builds have ucx as dependency, increase environment size


wm75:
### Solution to issue cannot be found in the documentation.

- [X] I checked the documentation.

### Issue

With the latest linux-64 openmpi builds *ucx* has started appearing as a dependency, instead of just a constraint:

```
$conda search openmpi --info -c conda-forge

...

openmpi 4.1.3 h846660c_102
--------------------------
file name   : openmpi-4.1.3-h846660c_102.tar.bz2
name        : openmpi
version     : 4.1.3
build       : h846660c_102
build number: 102
size        : 4.1 MB
license     : BSD-3-Clause
subdir      : linux-64
url         : https://conda.anaconda.org/conda-forge/linux-64/openmpi-4.1.3-h846660c_102.tar.bz2
md5         : b1498a2232331b919a91fe0c23d4704a
timestamp   : 2022-04-26 11:05:00 UTC
constraints : 
  - cudatoolkit  >= 10.2
  - ucx
dependencies: 
  - libgcc-ng >=10.3.0
  - libgfortran-ng
  - libgfortran5 >=10.3.0
  - libstdcxx-ng >=10.3.0
  - libzlib >=1.2.11,<1.3.0a0
  - mpi 1.0 openmpi
  - ucx >=1.12.1,<1.13.0a0
  - zlib >=1.2.11,<1.3.0a0


openmpi 4.1.3 h846660c_103
--------------------------
file name   : openmpi-4.1.3-h846660c_103.tar.bz2
name        : openmpi
version     : 4.1.3
build       : h846660c_103
build number: 103
size        : 4.2 MB
license     : BSD-3-Clause
subdir      : linux-64
url         : https://conda.anaconda.org/conda-forge/linux-64/openmpi-4.1.3-h846660c_103.tar.bz2
md5         : 3c58c19e8f53f54221dc26f6f0cef5c0
timestamp   : 2022-04-27 09:05:39 UTC
constraints : 
  - cudatoolkit  >= 10.2
  - ucx
dependencies: 
  - libgcc-ng >=10.3.0
  - libgfortran-ng
  - libgfortran5 >=10.3.0
  - libstdcxx-ng >=10.3.0
  - libzlib >=1.2.11,<1.3.0a0
  - mpi 1.0 openmpi
  - ucx >=1.12.1,<1.13.0a0
  - zlib >=1.2.11,<1.3.0a0


[...]

PR #175, opened 2024-09-21
Add UCC support

j34ni:
### Summary
- Added support for UCC (`build_with_ucc="--with-ucc=$PREFIX"`)
- Added addition `ompi_info` based tests

### Checklist


recipe/build-mpi.sh
```diff
@@ -75,27 +82,31 @@ fi
...
-            --enable-mca-dso \
             --enable-ipv6 \
             $build_with_ucx \
+            $build_with_ucc \
             $build_with_cuda \
     || (cat config.log; false)
 
-make -j"${CPU_COUNT:-1}"
+# Building with 16 CPUs
+make -j16
```
minrk, line 96: I don't think we should hardcode like this. These packages are built on CI, which doesn't have 16 cpus (at most 2, I think). Over subscribing like this often slows things down and greatly increases memory usage, sometimes to the point of crashing on low-resource CI. It should already be building with all available cpus by default, via CPU_COUNT. Is this not what you see?

minrk (commented):
Thanks for working on this!

recipe/build-mpi.sh
j34ni: Hi Min,
It was originally building with `make -j"${CPU_COUNT:-1}"` and that took ages, but we can obviously put that back

minrk: Were you building outside conda build? Conda build sets CPU_COUNT to the total number of cpus. If you are running this script outside, you can set CPU_COUNT=16, for example to build with 16 threads.

j34ni: Hi,

I was building on a VM, but that's Ok now: I set it back how it was, and I can see that on the Azure DevOps it builds in fact much faster than on my VM

Anyway, the issues I have now are for ppc64le (no ucc available) and osx, I am trying various things at the moment (although the syntax is very likely wrong)

j34ni: The build fails for linux_ppc64le: *RuntimeError: Could not get compatibility information for ucc package.  Is it one of your host dependencies?*

I tried `- ucc  # [linux and not ppc64le and not osx]` but that did not work: any idea what would be the correct syntax?

j34ni:
The build for ppc64le fails and I cannot understand why.
Here is the error message, can you tell what went wrong?
```
2024-09-23T07:45:44.9748755Z configure: exit 1
2024-09-23T07:45:44.9748969Z + false
2024-09-23T07:45:44.9749221Z Traceback (most recent call last):
2024-09-23T07:45:44.9749819Z   File "/opt/conda/lib/python3.10/site-packages/conda_build/build.py", line 1788, in bundle_conda
2024-09-23T07:45:44.9750151Z     utils.check_call_env(
2024-09-23T07:45:44.9750603Z   File "/opt/conda/lib/python3.10/site-packages/conda_build/utils.py", line 404, in check_call_env
2024-09-23T07:45:44.9751061Z     return _func_defaulting_env_to_os_environ("call", *popenargs, **kwargs)
2024-09-23T07:45:44.9752097Z   File "/opt/conda/lib/python3.10/site-packages/conda_build/utils.py", line 380, in _func_defaulting_env_to_os_environ
2024-09-23T07:45:44.9752714Z     raise subprocess.CalledProcessError(proc.returncode, _args)
2024-09-23T07:45:44.9753500Z subprocess.CalledProcessError: Command '['/usr/bin/bash', '-e', '/home/conda/feedstock_root/build_artifacts/openmpi-mpi_1727077155962/work/build-mpi.sh']' returned non-zero exit status 1.
2024-09-23T07:45:44.9753895Z 
2024-09-23T07:45:44.9754263Z The above exception was the direct cause of the following exception:
2024-09-23T07:45:44.9754508Z 
2024-09-23T07:45:44.9754834Z Traceback (most recent call last):
2024-09-23T07:45:44.9755303Z   File "/opt/conda/bin/conda-build", line 11, in <module>
2024-09-23T07:45:44.9755685Z     sys.exit(execute())
2024-09-23T07:45:44.9756229Z   File "/opt/conda/lib/python3.10/site-packages/conda_build/cli/main_build.py", line 589, in execute
2024-09-23T07:45:44.9756639Z     api.build(
2024-09-23T07:45:44.9757134Z   File "/opt/conda/lib/python3.10/site-packages/conda_build/api.py", line 209, in build
2024-09-23T07:45:44.9757537Z     return build_tree(
2024-09-23T07:45:44.9758072Z   File "/opt/conda/lib/python3.10/site-packages/conda_build/build.py", line 3712, in build_tree
2024-09-23T07:45:44.9758497Z     packages_from_this = build(
2024-09-23T07:45:44.9759044Z   File "/opt/conda/lib/python3.10/site-packages/conda_build/build.py", line 2765, in build
2024-09-23T07:45:44.9759492Z     newly_built_packages = bundlers[pkg_type](
2024-09-23T07:45:44.9760052Z   File "/opt/conda/lib/python3.10/site-packages/conda_build/build.py", line 1795, in bundle_conda
2024-09-23T07:45:44.9760515Z     raise BuildScriptException(str(exc), caused_by=exc) from exc
2024-09-23T07:45:44.9761301Z conda_build.exceptions.BuildScriptException: Command '['/usr/bin/bash', '-e', '/home/conda/feedstock_root/build_artifacts/openmpi-mpi_1727077155962/work/build-mpi.sh']' returned non-zero exit status 1.
2024-09-23T07:45:52.0075268Z 
2024-09-23T07:45:52.0477672Z ##[error]Bash exited with code '1'.
2024-09-23T07:45:52.0742477Z ##[section]Finishing: Run docker build
```

j34ni:
I spotted this in the log fir the build with ppc64le:
```
configure:125428: checking for sn/mmtimer.h
configure:125428: powerpc64le-conda-linux-gnu-cc -c -DNDEBUG -mcpu=power8 -mtune=power8 -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O3 -pipe -isystem /home/conda/feedstock_root/build_artifacts/openmpi-mpi_1727077155962/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold/include -fdebug-prefix-map=/home/conda/feedstock_root/build_artifacts/openmpi-mpi_1727077155962/work=/usr/local/src/conda/openmpi-5.0.5 -fdebug-prefix-map=/home/conda/feedstock_root/build_artifacts/openmpi-mpi_1727077155962/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold=/usr/local/src/conda-prefix -isystem /usr/local/cuda/targets/ppc64le-linux/include -finline-functions -DNDEBUG -D_FORTIFY_SOURCE=2 -O2 -isystem /home/conda/feedstock_root/build_artifacts/openmpi-mpi_1727077155962/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold/include -isystem /usr/local/cuda/targets/ppc64le-linux/include conftest.c >&5
conftest.c:693:10: fatal error: sn/mmtimer.h: No such file or directory
  693 | #include <sn/mmtimer.h>
      |          ^~~~~~~~~~~~~~
compilation terminated.
configure:125428: $? = 1
configure: failed program was:
| /* confdefs.h */
```

Could that be what makes the build fail?

recipe/meta.yaml
```diff
@@ -54,23 +55,56 @@ outputs:
...
         #- prrte
-        - libhwloc
+        - binutils  # [linux]
         - libevent
+        - libhwloc
+        - libmagic
         - libnl  # [linux]
+        - libnuma  # [linux]
+        - libudev1  # [linux]
+        - numactl  # [linux]
```
minrk, line 84: Are all of these new dependencies that are linked directly in order for this feature to work? That seems like a lot of direct dependencies to add.

recipe/meta.yaml
```diff
@@ -54,23 +55,56 @@ outputs:
...
+        - pkg-config  # [linux]
+        - rdma-core  # [linux]
+        - ucc  # [linux and not osx and not ppc64le]
+        - ucx  # [linux and not osx]
```
minrk, line 88: `linux and not osx` is the same as `linux`, no need to add `and not osx`

recipe/meta.yaml
```diff
@@ -101,6 +135,19 @@ outputs:
...
+      # Ensure the correct dependencies are available for testing
+      requires:      
+        - mpi 1.0 openmpi
+        - libevent
+        - libhwloc
+        - libnl  # [linux]
+        - libnuma  # [linux]
+        - libudev1  # [linux]
+        - numactl  # [linux]
+        - rdma-core  # [linux]
```
minrk, line 147: Each dependency added to the test env is indicating that it is a dependency that _users_ must specify explicitly in order for this to work. I suspect this should really be _only_ ucc, but maybe I'm wrong?

recipe/meta.yaml
```diff
@@ -54,23 +55,56 @@ outputs:
...
         #- prrte
+        - libevent
+        - libhwloc
+        - libnl  # [linux]
+        - libnuma  # [linux]
+        - libudev1  # [linux]
+        - numactl  # [linux]
+        - rdma-core  # [linux]
+        - ucc  # [linux and not osx and not ppc64le]
+        - ucx  # [linux and not osx]
```
minrk, line 103: ucx is specfically not a runtime dependency, and I _imagine_ ucc should be treated the same. Try removing _all_ of these new runtime dependencies, leaving only the `run_constrained` addition.

recipe/build-mpi.sh
```diff
@@ -33,15 +36,19 @@ if [[ "$target_platform" == osx-* ]]; then
     wrapper_ldflags='-Wl,-rpath,${libdir}'
 fi
 
-# UCX support
+# UCX/UCC support
 build_with_ucx=""
+build_with_ucc=""
 if [[ "$target_platform" == linux-* ]]; then
```
minrk, line 42: I think this is where you need a check for `${target_platform} != "linux-ppc64le"`

recipe/build-mpi.sh
```diff
@@ -75,27 +82,30 @@ fi
             --with-mpi-moduledir='${includedir}' \
             --with-wrapper-ldflags="${wrapper_ldflags}" \
             --with-sge \
+            --with-pmix=internal \
             --with-hwloc=$PREFIX \
             --with-libevent=$PREFIX \
             --with-zlib=$PREFIX \
-            --enable-mca-dso \
```
minrk, line 81: I think we need to make sure this option is still enabled, otherwise optional feature support becomes required at runtime

recipe/run_test.sh
```diff
@@ -27,7 +34,8 @@ if [[ $PKG_NAME == "openmpi" ]]; then
   $MPIEXEC --help
   $MPIEXEC -n 4 ./helloworld.sh
 
-  test -f $PREFIX/include/mpi.mod
+# test -f $PREFIX/include/mpi.mod
```
minrk, line 37: make sure to put this test back. If it fails, something is broken that needs to be fixed.

recipe/meta.yaml
```diff
@@ -54,23 +55,56 @@ outputs:
...
+        - libnuma  # [linux]
+        - libtool  # [unix]
+        - libudev1  # [linux]
         - make  # [unix]
+        - numactl  # [linux]
+        - pkg-config 
+        - rdma-core  # [linux]
+        - ucc  # [linux and not osx and not ppc64le]
+        - ucx  # [linux and not osx]
+        - zlib
```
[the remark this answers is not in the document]
j34ni: OK

recipe/meta.yaml
j34ni: OK

recipe/meta.yaml
j34ni: OK

recipe/meta.yaml
j34ni: OK

recipe/build-mpi.sh
j34ni: Restored it now

recipe/build-mpi.sh
j34ni: I think I got it working now

recipe/build-mpi.sh
j34ni: It works!

minrk:
Great work! This does seem to change ucx/ucc from optional to required dependencies. That means they shouldn't be present explicitly in `run` or `test` dependencies, they should either:

- be optional, and present in `ignore_run_exports`, or
- be required, and removed from `ignore_run_exports` (which puts the right pinning in `run`).

I think #102 is the motivation for making them optional, if we can keep going with that. Maybe with all of it working now, you can treat ucc the same as ucx and it will still work?

@leofeng can you look at this?

j34ni:
Thanks

I checked and it is not necessary to load UCX to use OpenMPI if there is no communications via interconnect involved (i.e., on a laptop or on a single node): it works without error or warning

However for use without CUDA support one has to do:
```
export OMPI_MCA_opal_warn_on_missing_libcuda=0
export OMPI_MCA_opal_cuda_support=0
export OMPI_MCA_mca_base_component_show_load_errors=0
```

recipe/meta.yaml
```diff
@@ -101,6 +110,11 @@ outputs:
         # host C compilers work fine (better) on mac
         - {{ compiler('c') }}
     test:
+      # Ensure the correct dependencies are available for testing
+      requires:      
+        - mpi 1.0 openmpi
```
leofang, line 115: This should not be needed, if so it's a bug in conda-build not in this recipe, we never had to add it. Let's keep it clean:
```suggestion
```

recipe/meta.yaml
```diff
@@ -101,6 +110,11 @@ outputs:
...
+        - ucc  # [linux and not ppc64le]
+        - ucx  # [linux and not ppc64le]
```
leofang, line 117: I may have missed something, but if you already added them to run (which is wrong, as noted above) why did you have to add them here again? This seems incorrect too.

leofang (changes requested):
> However for use without CUDA support one has to do:
> 
> ```
> export OMPI_MCA_opal_warn_on_missing_libcuda=0
> export OMPI_MCA_opal_cuda_support=0
> export OMPI_MCA_mca_base_component_show_load_errors=0
> ```

This is because you tweaked it wrong. We already set this in the provided config file. The idea is we build every possible support at build time, but not enable any by default. Users opt in for advanced features such as CUDA and UCX at process launch time. We do not want them having to opt out of any of such features.

recipe/build-mpi.sh
```diff
@@ -80,22 +96,26 @@ fi
...
             --enable-mca-dso \
             --enable-ipv6 \
+            $build_with_pmix \
             $build_with_ucx \
+            $build_with_ucc \
             $build_with_cuda \
     || (cat config.log; false)
 
-make -j"${CPU_COUNT:-1}"
+make -j"${CPU_COUNT:-1}"
```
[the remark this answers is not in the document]
j34ni: OK

recipe/build-mpi.sh
```diff
@@ -33,15 +38,26 @@ if [[ "$target_platform" == osx-* ]]; then
...
     build_with_ucx="--with-ucx=$PREFIX"
+    build_with_ucc="--with-ucc=$PREFIX"
+fi
+
+# PMIx support
+build_with_pmix=""
+if [[ "$target_platform" == linux-* && "$target_platform" != linux-ppc64le* ]]; then
+    echo "Build with PMIx support"
+    build_with_pmix="--with-pmix=internal"
 fi
```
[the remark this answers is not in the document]
j34ni: PMIx was added to support the use of OpenMPI on high-performance computing environments with Slurm, ensuring efficient process management and resource allocation
I tested it and it works well

recipe/build-mpi.sh
```diff
@@ -10,6 +10,11 @@ export FC=$(basename "$FC")
 # unset unused Fortran compiler variables
 unset FFLAGS F77 F90 F95
 
+# add -fallow-argument-mismatch to FCFLAGS
+if [[ "$target_platform" != linux-ppc64le ]]; then
+   export FCFLAGS="-fallow-argument-mismatch ${FCFLAGS}"
+fi
```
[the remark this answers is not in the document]
j34ni: The `export FCFLAGS="-fallow-argument-mismatch"` flag is used to prevent the GNU Fortran compiler (starting from version 10) from generating errors for argument mismatches in function and subroutine calls. This ensures compatibility with legacy Fortran code or codebases that do not strictly follow argument type matching, allowing for smoother compilation without interruptions.

recipe/meta.yaml
j34ni: OK, let's remove these lines

leofang:
@j34ni the last commit might have some problem. Rerendering failed and all CI configs were gone.

j34ni:
What happened to the `.ci_support` file?

j34ni:
Can you restore them?

leofang:
btw, before my review I see the following CI log. I'd like to point out two items:
- `pmix: internal`
- `OpenFabrics OFI Libfabric: no (not found)`

@j34ni it seems adding the additional packages to host (ex: pmix, libnuma, ...) is not doing what you want. Could you check if we should pass the build-time options differently so that Open MPI's build system can detect the new features that you need?
```
Open MPI configuration:
-----------------------
Version: 5.0.5
MPI Standard Version: 3.1
Build MPI C bindings: yes
Build MPI Fortran bindings: mpif.h, use mpi, use mpi_f08
Build MPI Java bindings (experimental): no
Build Open SHMEM support: yes
Debug build: no
Platform file: (none)
 
Miscellaneous
-----------------------
Atomics: GCC built-in style atomics
Fault Tolerance support: mpi
HTML docs and man pages: installing packaged docs
hwloc: external
libevent: external
Open UCC: yes
pmix: internal
PRRTE: internal
Threading Package: pthreads
 
Transports
-----------------------
Cisco usNIC: no
Cray uGNI (Gemini/Aries): no
Intel Omnipath (PSM2): no (not found)
Open UCX: yes
OpenFabrics OFI Libfabric: no (not found)
Portals4: no (not found)
Shared memory/copy in+copy out: yes
Shared memory/Linux CMA: yes
Shared memory/Linux KNEM: no
Shared memory/XPMEM: no
TCP: yes
 
Accelerators
-----------------------
CUDA support: yes
ROCm support: no
 
OMPIO File Systems
-----------------------
DDN Infinite Memory Engine: no
Generic Unix FS: yes
IBM Spectrum Scale/GPFS: no (not found)
Lustre: no (not found)
PVFS2/OrangeFS: no
```

leofang:
> Can you restore them?

You can just revert the commit and retry.

recipe/meta.yaml
```diff
@@ -54,23 +55,29 @@ outputs:
...
         #- openpmix
         #- prrte
-        - libhwloc
         - libevent
+        - libhwloc
         - libnl  # [linux]
+        - libnuma  # [linux and not ppc64le]
+        - libudev1  # [linux and not ppc64le]
+        - numactl  # [linux and not ppc64le]
+        - rdma-core  # [linux and not ppc64le]
```
leofang, line 69: Q: are `libnuma` and `rdma-core` really needed at build time? I see no linkage reported 
```
...
2024-09-24T12:11:46.3930863Z WARNING (openmpi): dso library package conda-forge/linux-64::libnuma==2.0.18=h4ab18f5_2 in requirements/run but it is not used (i.e. it is overdepending or perhaps statically linked? If that is what you want then add it to `build/ignore_run_exports`)
...
2024-09-24T12:11:46.3933755Z WARNING (openmpi): run-exports library package conda-forge/linux-64::rdma-core==53.0=he02047a_0 in requirements/run but it is not used (i.e. it is overdepending or perhaps statically linked? If that is what you want then add it to `build/ignore_run_exports`)
...
```

cc @dalcinl @pentschev @marcinz in case you can shed some light here

recipe/build-mpi.sh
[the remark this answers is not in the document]
leofang: I meant why we didn't add it earlier :) 

But, see also https://github.com/conda-forge/openmpi-feedstock/pull/175#issuecomment-2374252917, I don't think this works yet

leofang:

recipe/meta.yaml
```diff
@@ -120,6 +128,11 @@ outputs:
         # host C compilers work fine (better) on mac
         - {{ compiler('cxx') }}
     test:
+      # Ensure the correct dependencies are available for testing
+      requires:
+        - mpi 1.0 openmpi
+        - ucc  # [linux and not ppc64le]
+        - ucx  # [linux and not ppc64le]
```
leofang, line 135:
```suggestion
```

recipe/meta.yaml
```diff
@@ -138,6 +151,10 @@ outputs:
         - {{ pin_subpackage('openmpi', exact=True) }}
         - {{ compiler('fortran') }}
     test:
+      requires:
+        - mpi 1.0 openmpi
+        - ucc  # [linux and not ppc64le]
+        - ucx  # [linux and not ppc64le]
```
leofang, line 157:
```suggestion
```

recipe/meta.yaml
```diff
@@ -101,6 +108,7 @@ outputs:
         # host C compilers work fine (better) on mac
         - {{ compiler('c') }}
     test:
+      # Ensure the correct dependencies are available for testing
```
leofang, line 111:
```suggestion
```

recipe/meta.yaml
pentschev: I wonder if those are requirements that are only needed by UCX, I don't know anything about them in OpenMPI though. I know [libnuma's dependency was removed from UCX quite some time ago](https://github.com/openucx/ucx/pull/8698), and rdma-core is used to provide InfiniBand support, which I don't know if OpenMPI has native support for (i.e., without depending on UCX).

leofang:
```
  File "/tmp/tmpnjif5hwu/openmpi-feedstock/recipe/meta.yaml", line 83, in top-level template code
    # Ensure a consistent CUDA environment
^^^^^^^^^^^^^^^^^^^^^
jinja2.exceptions.UndefinedError: 'CONDA_BUILD_SYSROOT' is undefined
```
conda-{smithy,build} are being naughty lately... have anyone seen this error elsewhere?

leofang:

j34ni:
@leofang  I think I messed up, sorry about that. I have reverted for now a couple of commits and will go again through the comments

recipe/meta.yaml
j34ni: These dependencies may not be mandatory at build time for our core functionality.
They support specific advanced features such as NUMA awareness and RDMA capabilities. 
I guess that users interested in these optimizations can manually load the respective libraries if required.

j34ni:
I am trying to use the current build (with all the changes you suggested) on a HPC to run the OSU Micro Benchmarks 7.4 and there must be something I do not understand because it fails.
The exact same thing worked with my previous build.
How do you suggest to *activate* UCX with the current build?

leofang:
>> conda-{smithy,build} are being naughty lately... have anyone seen this error elsewhere?
> 
> @leofang I think I messed up, sorry about that. 

No, not your fault at all. The rerendering functionality seems to be down at least today, which affects both offline rerendering and CI services. The core team is investigating: https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/350#issuecomment-2375500712.

We need the rerendering capability to be back before we can rebuild the package for local testing, because rerendering would update the `.ci_support` config to the newly added packages, e.g.
https://github.com/conda-forge/openmpi-feedstock/blob/954a28450bf11f01853995a3fd64f3b16e18b6ff/.ci_support/linux_64_c_compiler_version11cuda_compilernvcccuda_compiler_version11.8cxx_compiler_version11fortran_compiler_version11mpi_typeconda.yaml#L37-L38
and the config is then fed into conda-build.

> The exact same thing worked with my previous build.
> How do you suggest to _activate_ UCX with the current build?

The logic is reverted, so now you need a way to activate the UCX/UCC support if you need them. For UCX we've figured it out here: https://github.com/conda-forge/openmpi-feedstock/blob/main/recipe/post-link-ucx.sh Could you kindly take a look and see how to generalize it to UCX, please?

leofang:
> there must be something I do not understand because it fails.

Failed in what way?

j34ni:
I removed the unnecessary dependencies for the host and run

As far as I understand if OpenMPI is configured/built `--with-ucx=... --with-ucc=...` then it will use UCX/UCC at run time if it is available, and will not complain if it is not used and not available.

j34ni:
@leofang

UCC added support for the MPI_Scatter() and MPI_Iscatter() collective, and things like `--mca coll_tuned_use_dynamic_rules` can be used to tune the collective component, but I can see nothing to turn it on/off

I guess that what you had for UCX could do for both then

recipe/meta.yaml
leofang: @j34ni would you mind updating the PR title/description to reflect the above discussion and the latest changes?

leofang:
> (...) but I can see nothing to turn it on/off
> 
> I guess that what you had for UCX could do for both then

@j34ni could you try if `--mca coll ucc` (enable) and `--mca coll ^ucc` (disable) would make a difference? If so we should
- update the `conf` file to add `^ucc` (disable by default)
- figure out how to enable/disable it through env vars 
- update the post link script to hint users about how to enable it

A few more words: We rely on Open MPI's MCA mechanism to disable/enable a component. The `ucc` component is an available option of the `coll` framework. It is important that we understand the inner working even from the packaging perspective, because it affects how we build a package, set its default expectation, and enable advanced use cases.
- https://docs.open-mpi.org/en/v5.0.x/mca.html#frameworks
- https://docs.open-mpi.org/en/v5.0.x/tuning-apps/coll-tuned.html

recipe/meta.yaml
j34ni: Is that better?

j34ni:
I think I got it now and we can enable or disable UCC easily:

To disable it: `export OMPI_MCA_coll_ucc_enable=0`

To enable it: `export OMPI_MCA_coll_ucc_enable=1`

recipe/build-mpi.sh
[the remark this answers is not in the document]
leofang: @j34ni The ask was to investigate why you added `pmix` in the host section but we still see it shown as "internal" instead of "yes" (which would mean the conda-forge `pmix` is detected and to be used at build time), not to change to `--with-pmix=internal` here. 

As the name implied, Open MPI already has a copy of PMIx and so when the build system cannot detect any installed in the user environment, it'd use the private copy. See e.g. the build log from the main branch: https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=991597&view=logs&j=d0d954b5-f111-5dc4-4d76-03b6c9d0cf7e&t=779a0c34-7b19-5ccb-5da7-fe4579c91392&l=5814

recipe/meta.yaml
leofang: Yeah thanks!

recipe/build-mpi.sh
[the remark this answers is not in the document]
leofang: Thanks, @j34ni. A follow-up question: This flag is only used at build time, but Open MPI was already built fine, so what issues does this solve? I don't think I get it.

recipe/build-mpi.sh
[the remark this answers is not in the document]
j34ni: This is because I originally wanted to use an external installation of PMIx (to offer more flexibility, as for libevent and hwloc) and hence to configure with `--with-pmix=$PREFIX`, however the https://github.com/openpmix/pmix-feedstock is not ready yet so I temporarily used `--with-pmix=internal` as a placeholder (rather than not even mentioning it at all and letting people doing an `ompi_info` think that it was not supported)

recipe/run_test.sh
```diff
@@ -7,14 +7,24 @@ pushd "tests"
 
 if [[ $PKG_NAME == "openmpi" ]]; then
 
-  if [[ -n "$(conda list | grep ucx)" ]]; then
-    echo "Improper UCX dependency!"
-    exit 1
```
leofang, line 12: We need to keep the old tests. The purpose is different from the new ones that you added:
- `conda list` tests: ensure the optional dependencies are not installed by default
- `ompi_info` tests: ensure the support for optional dependencies was built

recipe/run_test.sh
```diff
@@ -7,14 +7,24 @@ pushd "tests"
...
+  if [[ "$target_platform" == linux-64 ]]; then
+     if [[ -z "$(ompi_info | grep ucx)" ]]; then
+        echo "OpenMPI configured without UCX support!"
+        exit 1
+     fi
   fi
 
-  if [[ -n "$(conda list | grep cuda-version)" ]]; then
-    echo "Improper CUDA dependency!"
-    exit 1
```
leofang, line 17: ditto

recipe/run_test.sh
```diff
@@ -84,3 +94,4 @@ if [[ $PKG_NAME == "openmpi-mpifort" ]]; then
 fi
 
 popd
+
```
leofang, line 97: nit: extra empty line
```suggestion
```

recipe/build-mpi.sh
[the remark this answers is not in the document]
j34ni: This is because if it is not here there can be arguments mismatches later in codes compiled with mpifort (that happens with CESM typically), so I added it to prevent this kind of errors

recipe/post-link-ucc.sh
```diff
@@ -0,0 +1,13 @@
...
+
+cat << EOF >> $PREFIX/.messages.txt
+
+On Linux, Open MPI is built with UCC support but it is disabled by default.
+To enable it, first install UCC (conda install -c conda-forge ucc).
+Afterwards, set the environment variables
+OMPI_MCA_coll_ucc_enable=1
+before launching your MPI processes.
+Equivalently, you can set the MCA parameters in the command line:
+mpiexec --mca coll_ucc_enable 1 ...
```
leofang, line 11: I think this is correct, thank you for checking 👍 

I have mixed feeling in this. While following the MCA convention we can just do
```
OMPI_MCA_coll=ucc
```
or
```
mpiexec --mca coll ucc
```
to enable the UCC support, this is surprisingly not taught in, e.g., UCC's documentation:
https://github.com/openucx/ucc/blob/master/docs/user_guide.md#getting-started
so I guess we should follow what the UCC team suggests, given lack of `coll` documentation in Open MPI... (cc: @jsquyres for vis)

leofang:

minrk:
The main reason ucx was excluded as a hard dependency was to save space because it resulted in hundreds of megabytes of extra packages due to cuda stuff. That's not true anymore since #119, so would it simplify things to make ucc/ucx a plain conda dependency? I don't entirely follow the reasoning behind ucx still being optional when it seems to only save ~20MB. The functionality is still optional, right?

leofang:

recipe/build-mpi.sh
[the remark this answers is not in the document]
leofang: Oh I did not realize there's no `pmix` on conda-forge. Note that we cannot build a package that would depend on packages from another channel.

> so I temporarily used `--with-pmix=internal` as a placeholder

In-line comments would not help issue tracking (and there's no comment added here to describe that this is for placeholding). I suggest opening an issue to track the need, and undo what's done here:
```suggestion
```

> letting people doing an `ompi_info` think that it was not supported

As described above, it is supported. Lack of `--with-pmix=internal` in `Configure command line` does not mean PMIx was not used at build time. See the linked build log from the main branch.

recipe/build-mpi.sh
```diff
@@ -80,7 +96,9 @@ fi
             --with-zlib=$PREFIX \
             --enable-mca-dso \
             --enable-ipv6 \
+            $build_with_pmix \
```
leofang, line 99:
```suggestion
```

recipe/build-mpi.sh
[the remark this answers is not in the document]
leofang: > I suggest opening an issue to track the need

xref: #178

recipe/build-mpi.sh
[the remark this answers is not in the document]
leofang: Does the build-time `FCFLAGS` get recorded in `mpifort`? If not, I would think this is added to the wrong place, and it should be added to the activation script of `openmpi-mpifort` instead. Furthermore, by doing so it would cause a global change in whoever using `openmpi-mpifort`, and I would be hesitant in propagating to downstream projects/users. @minrk @dalcinl thoughts?

leofang:
> No, not your fault at all. The rerendering functionality seems to be down at least today, which affects both offline rerendering and CI services. The core team is investigating: [conda-forge/conda-forge-ci-setup-feedstock#350 (comment)](https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/350#issuecomment-2375500712).

To close the loop, it turns out the rerendering failed for another reason: https://github.com/conda-forge/openmpi-feedstock/pull/177#discussion_r1777665637 which is now fixed.

leofang:
> I don't entirely follow the reasoning behind ucx still being optional when it seems to only save ~20MB. The functionality is still optional, right?

My preference is if something is optional, we do not list them in `run`. In addition to extra bandwidth and storage usage, adding more dependencies could also run into dependency resolving issues. Right now this PR presents a minimal change to the recipe and I like it that way. Maybe we can move this discussion to a separate issue and try to get this merged once the last question on `FCFLAG` is resolved and the tests pass?

minrk:
Makes sense!

recipe/build-mpi.sh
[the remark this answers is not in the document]
minrk: CFLAGS/etc do not get recorded in the wrappers, that would be in `--wrapper-cflags` args, as we currently have for `--wrapper-ldflags` ([docs](https://docs.open-mpi.org/en/v5.0.x/installing-open-mpi/compilers-and-flags.html)). This arg here only affects the build of openmpi itself, so if openmpi itself doesn't need it, we can leave it out.

minrk:
Thanks, this LGTM now.

recipe/run_test.sh
j34ni: I find it much more relevant to verify that OpenMPI has effectively been built with the desired features (here UCX and CUDA).
This approach is more comprehensive because it goes beyond verifying the mere presence of dependencies and checks the actual runtime configuration and capabilities of the package, providing stronger assurance that the package will function correctly in the user's environment.

minrk:
Thanks @j34ni and thanks @leofang for reviewing!

j34ni:
Thanks to you guys
________________________________

leofang:
Thanks @j34ni @minrk @pentschev!

leofang:
FYI as a follow-up I'm adding UCC to the global pinning https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/6484

leofang:
> Maybe we can move this discussion to a separate issue

See #181.

PR #188, opened 2024-11-18
use external libpmix

### LINKED ISSUE #178: Support `pmix` from conda-forge?
xref: https://github.com/conda-forge/openmpi-feedstock/pull/175/files#r1777831382

minrk:
This unbundles libfabric (ready) and libpmix, prrte (waiting on https://github.com/conda-forge/staged-recipes/pull/28225).

I've run this build locally on osx-arm64 and everything works. Draft because pmix/prrte are not published yet.

closes #178

### DIFF
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -11,27 +11,27 @@ jobs:
       ? linux_64_c_compiler_version11cuda_compilernvcccuda_compiler_version11.8cxx_compiler_version11fortran_compiler_version11mpi_typeconda
       : CONFIG: linux_64_c_compiler_version11cuda_compilernvcccuda_compiler_version11.8cxx_compiler_version11fortran_compiler_version11mpi_typeconda
         UPLOAD_PACKAGES: 'True'
-        DOCKER_IMAGE: quay.io/condaforge/linux-anvil-cuda:11.8
+        DOCKER_IMAGE: quay.io/condaforge/linux-anvil-x86_64-cuda11.8:cos7
       ? linux_64_c_compiler_version11cuda_compilernvcccuda_compiler_version11.8cxx_compiler_version11fortran_compiler_version11mpi_typeexternal
       : CONFIG: linux_64_c_compiler_version11cuda_compilernvcccuda_compiler_version11.8cxx_compiler_version11fortran_compiler_version11mpi_typeexternal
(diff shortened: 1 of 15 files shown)


dalcinl:
@minrk I would ask for advise upstream to the Open MPI team about the feasibility of this PR. Currently, the `main` branch of the open-mpi/ompi is using a custom fork of prrte. Moreover, this fork has diverged to the point where things are hard to patch https://github.com/open-mpi/ompi/issues/12929#issuecomment-2488612537. The open-mpi/ompi@v5.0.x branch is fine, they are using upstream PMIx/PRRTE repos, but I do not really know what are the version pinning rules. I'm a little worried  that once you merge this PR, things may eventually break in upcoming releases.

@hppritcha Could you please drop us a comment about how should we handle external PMIx/PRRTE packages with respect to our Open MPI builds? Is there any official place stating the exact versions or PMIx/PRRTE that Open MPI requires?

minrk:
Yeah, I figured it would be synced to an upstream release by the time openmpi is ready for its own release, but if that's not the case and openmpi expects to carry its own unreleased forks, then we probably shouldn't do this. It would be good to have confirmation of how openmpi and prrte releases relate.

Judging by the [PR that introduced the fork](https://github.com/open-mpi/ompi/pull/12449) it looks like the plan going forward is to keep the fork divergent and not maintain compatibility with actual PRRTE.

FWIW, openpmix itself is still tracking upstream, not a fork. So it appears to be only PRRTE that's affected. In which case, we can still do this for pmix, but keep the PRRTE bundled. Or pull PRRTE out for openmpi 5 and switch back to bundled PRRTE for openmpi 6. Or just not do this altogether, which is honestly okay with me, too.

hppritcha:
For Open MPI 5.0.x releases, we support use of an external PRRte against which Open MPI 5.0.x releases can be built.

The minimum version of PMIx we support for Open MPI 5.0.x is 4.2.0.
The minimum version of PRRTe we support for Open MPI 5.0.x is 3.0.0.

See https://docs.open-mpi.org/en/v5.0.x/installing-open-mpi/required-support-libraries.html for more details about PMIx/PRRTe dependency info.  

Note that PRRTe has its own dependencies on PMIx versions that need to be kept in mind - https://docs.prrte.org/en/latest/install.html

minrk:
Thanks! Is it right that starting with openmpi 6, it is not planned to work with public prrte releases anymore? Is the same true for pmix?

PR #192, opened 2025-03-06
openmpi 4.1.7

dalcinl:
Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [x] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

Issue #110, opened 2022-06-04
Support {{ compiler('mpi') }}


isuruf:
### Comment:

This way, we can setup cross compiling env things like OPAL_PREFIX and build_env compiler.
Also, we can remove hacks like https://github.com/conda-forge/conda-smithy/blob/52db47bc3064f2471c1c8927c02d1149014557fa/conda_smithy/configure_feedstock.py#L280

cc @minrk, @beckermr

leofang:
Sounds like an alternative solution to #91 to me? (xref: https://github.com/conda-forge/conda-forge.github.io/pull/1752)

jakirkham:
cc @dalcinl

dalcinl:
Perhaps it should be `compiler('mpicc')`, `compiler('mpicxx')`, `compiler('mpifort')`?
Recipes may need only one, or two, or all of them. Depending on the actual semantics of `compiler('mpi')`, my comment may be irrelevant. Perhaps the idea is that `compiler('mpi')` alone is meaningless and should error, unless something like `compiler('c')`  is also present.

Or perhaps the `compiler(...)` thing could be extended such that it could be used the following way:

```yaml
{{ compiler('c', 'mpi') }}
{{ compiler('cxx', 'mpi') }}
{{ compiler('fortran', 'mpi') }}
```

Beside these minor details about defining actual semantics, this is a great idea!

PR #194, opened 2025-04-15
Open up compiler version bound for wrappers

### LINKED ISSUE #193: Compiler wrappers require gcc <= 12
The compiler wrapper packages openmpi-mpicc, openmpi-mpicxx, and openmpi-mpifort depend on a particular major version of gcc_linux-64 (at least on GNU/Linux, which I'm using), with newer versions not allowed.

The newest major version supported by the packages currently available is 12, meaning that gcc 13 (which is the default in conda-forge now) and gcc 14 cannot be used to compile Conda packages against openmpi.

Since gcc_linux-64 is just the compiler activation script, it doesn't seem like this restriction is needed? It seems to be generated by the dependency on `compiler('c')` in the meta.yaml, so maybe changing the version constraint isn't so easy. But perhaps some packages could be added that work with newer compilers?

LourensVeen:
Checklist

Closes #193.

This implements the fix proposed in the discussion for #193, setting the present compiler version as the minimum and removing the upper bound.

I'd love a 4.x series build with this fix too, for my current project 5.x seems to be working fine, but I've run into [issues](https://github.com/open-mpi/ompi/issues/12967) with 5.x in another context. I guess I'm asking for #192 really, but that's going a bit beyond my conda-fu I fear.


LourensVeen:

dalcinl (approved):

dalcinl:
@LourensVeen Any chance that something like `{{ compiler('c').replace(' ', '>=', 1) }}` would work?

LourensVeen:
Can't hurt to try :)

recipe/meta.yaml
```diff
@@ -110,7 +111,7 @@ outputs:
       run:
         - {{ pin_subpackage('openmpi', exact=True) }}
         # host C compilers work fine (better) on mac
-        - {{ compiler('cxx') }}
+        - {{ compiler('cxx').replace(' ', '>=', 1) }}
```
dalcinl, line 114:
```suggestion
        - {{ compiler('cxx') | replace(' ', '>=', 1) }}
```

LourensVeen: I figured I'd try both syntaxes, since I have no clue and you seemed unsure.

dalcinl: Indeed, I was just guessing... :sweat_smile:

LourensVeen:
Looks like it's working now, except that one of the tests failed because a download failed.

Is there any way I could access the built packages to see if the dependency is now actually what we want?

dalcinl:
You can store artifacts using `azure/store_build_artifacts` in `conda-config.yml`.
https://conda-forge.org/docs/maintainer/conda_forge_yml/#azure

LourensVeen:
@conda-forge-admin please rerender

LourensVeen:
Okay, I have stored and downloaded the alien artifacts, then installed them locally with dependencies from conda-forge, and I've compiled and run C, C++ and Fortran test programs with gcc 11.4.0, 12.4.0, 13.3.0, and 14.2.0, which works fine at least on linux-64.

The dependencies now look like this:
```
openmpi-mpicc-5.0.7-hc43e4ee_101/info$ cat index.json 
{
  "arch": "x86_64",
  "build": "hc43e4ee_101",
  "build_number": 101,
  "depends": [
    "__glibc >=2.17,<3.0.a0",
    "gcc_linux-64 >=11.*",
    "openmpi 5.0.7 hb85ec53_101"
  ],
  "license": "BSD-3-Clause",
  "license_family": "BSD",
  "name": "openmpi-mpicc",
  "platform": "linux",
  "subdir": "linux-64",
  "timestamp": 1744797675375,
  "version": "5.0.7"
}
```
which is what we intended.

There are still a couple of build failures due to network timeouts, those seem to be fairly common. Do I keep retrying until it passes?

LourensVeen:
@conda-forge admin, please restart ci

conda-forge.yml
```diff
@@ -1,4 +1,5 @@
 azure:
+  store_build_artifacts: true
```
dalcinl, line 2:
```suggestion
  store_build_artifacts: false
```

dalcinl:
@LourensVeen Are we all set? Should I merge once CI finishes?

leofang:

LourensVeen:
I think this is ready to go.

It's a bit scary because I'm not used to working on stuff that's this widely used, but it works as far as I can test it, passes all the built-in tests as well, and has a positive review, so let's go for it :smile:

LourensVeen:
Ah, I never realised that the actual `mpicc` command was in the `openmpi` package, not `openmpi-mpicc`. I figured the idea was that a package could have a dependency on `openmpi` and only use `openmpi-mpicc` while building. And that I would need `openmpi-mpicc` to build anything. Oh well :smile:.

LourensVeen:
There's still an open issue (#110) about `compiler('mpi')` actually.

Feeding MPI support into a build system isn't as standardised as $CC though. The $AX_MPI macro for autoconf uses MPICC and CMake has its own different set of environment variables. Setting $CC may work if the build system is just a Makefile, but with something more complicated it could also mislead it into thinking there's no MPI. I've found just using `mpicc`, `mpicxx` and `mpifort` pretty reliable, I think that works across MPICH, OpenMPI, IMPI and Cray MPI.

For background, I'm working to get [AMUSE](https://github.com/amusecode/amuse) into conda-forge. This is an astrophysics toolbox with some 60 codes in it that do a variety of calculations, written in C, C++, Fortran and CUDA, with MPI and OpenMP and OpenCL, and of varying vintage and quality, all wrapped in Python. In short, the perfect packaging problem :smile:.

Of course none of it was designed with packaging or binary distribution in mind, but it's also old and crufty and hard to compile, making packaging more important. On the other hand, the codes aren't always written to be runtime configurable either, and the retrofit of that isn't always complete, so sometimes you want to modify the code, which can be tricky if, like many of the users, you don't really know what a compiler is.

So I'm trying to have the users set up a more or less standardised build environment using conda-forge which then also matches conda-forge's build environment closely, so that mixing binary packages from conda-forge with locally-built packages for some of the codes can work, and also so that any contributions from users are less likely to build the conda-forge builds. When I get around to submitting some feedstocks that is.

For some of these codes setting CC to the MPI compiler is just what's needed, but most of them need something more complicated, and I fear it's probably similar across the computational science/HPC domain...

minrk:
Yeah, that's why I don't really think `compiler("mpi")` or setting `CC=mpicc` is really worth doing. There are too many ways to use mpi compiler wrappers, and it's far easier to do the right thing in a per-package way. `openmpi` has everything and there's no real need for any more packages.

The only real problem `openmpi-mpicxx` was meant to solve is that `clangxx_osx-64` was hard for users to remember and type (they were never meant to be used in recipes, where this already isn't a problem), and `cxx-compiler` already solves the same problem without needing a new package.

PR #223, opened 2026-08-20
Fix indirect dependency on CUDA

luhenry:
- **Fix indirect dependency on CUDA**
- **Fix linter**
- **MNT: Re-rendered with conda-smithy 2026.8.9 and conda-forge-pinning 2026.08.20.13.36.35**

Checklist
* [x] Reset the build number to `0` (if the version changed)


luhenry:
Relates to https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/8880. It will be necessary in the context of [the linux-riscv64 migration](https://conda-forge.org/status/migration/?name=supportlinuxriscv64platform) to remove dependency on (yet) unavailable CUDA toolchain on RISC-V

dalcinl:
@luhenry https://github.com/conda-forge/openmpi-feedstock/blob/main/recipe/run_test.sh#L29

recipe/run_test.sh
```diff
@@ -26,7 +26,7 @@ if [[ $PKG_NAME == "openmpi" ]]; then
     exit 1
   fi
 
-  if [[ "$target_platform" == linux-* ]]; then
+  if [[ "$target_platform" == linux-64 || "$target_platform" == linux-aarch64 ]]; then
```
dalcinl, line 29: @beckermr I do not know how to do better. Is the `cuda_compiler_version` somehow available when tests run? 

What we should really be doing here is the same as in the build script:

```bash
  cuda_version="${cuda_compiler_version:-}"
  if [[ ! -z "$cuda_version" && "$cuda_version" != "None" ]]; then
```

luhenry: We have been using this explicit `[[ "$target_platform" == linux-64 || "$target_platform" == linux-aarch64 ]]` elsewhere for CUDA checks (https://github.com/conda-forge/mpich-feedstock/pull/140/changes#diff-eba974640f1593934e24faefbf76a6109801a3d64bd40dc4fedcd621762da54eR62)

dalcinl: OK, fair enough.... I'm truly pedantic, sorry...

luhenry: There should indeed be a better way so that we don’t have to go through all the feedstock whenever CUDA brings support to riscv64 or some other platforms.

